Issue #35, opened 2019-04-24
Issue ingesting blocks with block_height as -1


garethtdavies:
After following installation instructions when starting zcashd with -reindex I get the following error:

```
INFO[2019-04-23T19:11:05-07:00] Listening to 0mq on 127.0.0.1:28332           app=zmqclient
INFO[2019-04-23T19:11:38-07:00] new block                                     app=zmqclient block_hash=382c4a332661c7ed0671f32a34d724619f086c61873bce7c99859dd9920aa605block_height=0 block_version=4 error="<nil>" sapling=false seqnum=0 tx_count=1
INFO[2019-04-23T19:11:38-07:00] new block                                     app=zmqclient block_hash=238d665a062b9007836a7d8f968ba2f3847af5f542733389a952cf9b86795502block_height=-1 block_version=4 error="<nil>" sapling=false seqnum=1 tx_count=1
ERROR[2019-04-23T19:11:38-07:00] new block                                     app=zmqclient block_hash=181ac2d6dc0ae73a82d7db472ac2879a7c24d8b12e5f73efc33a55549ea4f100 block_height=-1 block_version=4 error="storing compact block -1: UNIQUE constraint failed: blocks.block_height" sapling=false seqnum=2 tx_count=1
ERROR[2019-04-23T19:11:46-07:00] error on msg recv                             app=zmqclient error="interrupted system call"
```

In addition, viewing the sqlite database for the `blocks` table I see the following for `block_height` 0 and the `compact encoding` field:

```
 8,J3&aq*4$ala;|ْ
 
```

Running go 1.12.2 and zcashd v2.0.4 on Ubuntu 19.0.4.

garethtdavies:
This was a missing dependency. Ensured everything from https://github.com/zcash-hackworks/lightwalletd-z-cash/blob/master/Dockerfiles/lightwalletd/Dockerfile was installed and this cured it.

garethtdavies:
Actually, the above fix was coincidental and the underlying issue remains. 

I have found that for blocks 1-16 this line always returns -1 https://github.com/zcash-hackworks/lightwalletd/blob/master/parser/block.go#L71 so if you ingest a couple of blocks 1-16 then you'll get the `UNIQUE constraint failed` error.

If you don't ingest any blocks 1-16 then everything works as it should.

str4d:
[...]

Issue #32, opened 2019-04-16
Don't use block height as a primary key


gtank:
It's likely causing failures during reorgs or other transient failures that result in a rebroadcast. Found by @ianamunoz and @gmale.

https://github.com/zcash-hackworks/lightwalletd/blob/master/storage/sqlite3.go#L30

gmale:
This basically boils down to "implement re-orgs" which is a feature we knowingly punted on, as a trade-off.

gmale:
Related to and possibly a duplicate of #31

mms710:
Relates to https://github.com/zcash/zcash-android-wallet-sdk/issues/23

mdr0id:
No longer relevant due to updates to mentioned topics

PR #37, opened 2019-05-01
Use REPLACE on db row creation for initial reorg logic

mdr0id:
This PR is a naive solution(intended to partially address #32, #35  short term) to the existing reorgs issue that causes lightwalletd to abruptly/randomly stop.

`INSERT` cannot handle reorg block row creation, due to UNIQUE constraints on the row data.

`REPLACE` will remove and replace the row entry causing the unique constraint. 

My initial thoughts were lightwalletd can replace duplicate(containing same height and or hash) blocks, and assume the last received duplicate block will contain the chain-valid block. 

I need to go test this more, but thoughts/ideas are welcome.

(Disclaimer; I'm a Golang newb and just started with this portion of the stack a few days ago 😄 )

### DIFF
--- a/storage/sqlite3.go
+++ b/storage/sqlite3.go
@@ -123,6 +123,7 @@ func GetBlockRange(ctx context.Context, db *sql.DB, blockOut chan<- []byte, errO
 
 func StoreBlock(conn *sql.DB, height int, hash string, sapling bool, encoded []byte) error {
 	insertBlock := "REPLACE INTO blocks (block_height, block_hash, sapling, compact_encoding) values (?, ?, ?, ?)"
+
 	tx, err := conn.Begin()
 	if err != nil {
 		return errors.Wrap(err, fmt.Sprintf("creating db tx %d", height))
--- a/storage/sqlite3.go
+++ b/storage/sqlite3.go
@@ -122,8 +122,7 @@ func GetBlockRange(ctx context.Context, db *sql.DB, blockOut chan<- []byte, errO
 }
 
 func StoreBlock(conn *sql.DB, height int, hash string, sapling bool, encoded []byte) error {
-	insertBlock := "INSERT INTO blocks (block_height, block_hash, sapling, compact_encoding) values (?, ?, ?, ?)"
-
+	insertBlock := "REPLACE INTO blocks (block_height, block_hash, sapling, compact_encoding) values (?, ?, ?, ?)"
 	tx, err := conn.Begin()
 	if err != nil {
 		return errors.Wrap(err, fmt.Sprintf("creating db tx %d", height))


garethtdavies:
Testing this and it synced up to the current testnet height without issue. This seems like a great short term fix at the very least to aid with testing. 

One issue with this would be I don't think you can rely on the sequence to be correct as you can see from this screenshot of my sync it's all over the place so they might get inserted in the wrong order (see `seqnum`).

zcash-sync

mdr0id:
Your output is rather interesting. On my local machine the block height and seqnum appear to be incrementing as expected:

image

garethtdavies:
Are you also downloading the chain as I'm just resyncing an existing previously synced node so I guess it's churning through them quickly and it's queuing them?

mdr0id:
The node on this rig is pretty much synced(99%) on testnet and just ran with `--reindex`

garethtdavies:
Rather fortuitously there was a reorg at 482212 so I can confirm this did behave as expected.

lightwalletd:

```
INFO[2019-05-01T18:23:23-07:00] new block                                     app=zmqclient block_hash=adfc10edbd5d984f7c26045ff69f1a0f555d3f432f4c9758f69432324fd70900block_height=482212 block_version=4 error="<nil>" sapling=false seqnum=482217 tx_count=1
INFO[2019-05-01T18:23:51-07:00] new block                                     app=zmqclient block_hash=237246594d58d3f70fe7e44f85e527a8d6a79c82365073940845e10f6a630400block_height=482212 block_version=4 error="<nil>" sapling=false seqnum=482218 tx_count=1
```

zcashd logs:

```
May 01 18:23:24 pi-zcash-testnet debug.log:  2019-05-02 01:23:23 UpdateTip: new best=0009d74f323294f658974c2f433f5d550f1a9ff65f04267c4f985dbded10fcad  height=482212  log2_work=36.627033  tx=661546  date=2019-05-02 01:23:21 progress=1.000000  cache=0.1MiB(234tx)
May 01 18:23:51 pi-zcash-testnet debug.log:  2019-05-02 01:23:51 UpdateTip: new best=0004636a0fe1450894735036829ca7d6a827e5854fe4e70ff7d3584d59467223  height=482212  log2_work=36.627033  tx=661546  date=2019-05-02 01:23:22 progress=1.000000  cache=0.1MiB(234tx)
```

garethtdavies:
> The node on this rig is pretty much synced(99%) on testnet and just ran with `--reindex`

Similar. I'm on a pretty high-end machine i9-9900k and 64GB of RAM so it may be broadcasting these during a reindex faster than it can ingest them. Original point remains though that it probably isn't wise to solely rely on the order received as this seemingly isn't guaranteed.

mdr0id:
Interesting. 

Agreed, I figured this would fall down for reorg cases unfortunately .  Ever onward to see what can be done where 🤓

storage/sqlite3.go
```diff
@@ -122,8 +122,7 @@ func GetBlockRange(ctx context.Context, db *sql.DB, blockOut chan<- []byte, errO
 }
 
 func StoreBlock(conn *sql.DB, height int, hash string, sapling bool, encoded []byte) error {
-	insertBlock := "INSERT INTO blocks (block_height, block_hash, sapling, compact_encoding) values (?, ?, ?, ?)"
-
```
str4d: Nit: leave the blank line.

str4d: Pushed a commit myself to address the nit. Now the GitHub change view correctly shows that the only change was `s/INSERT/REPLACE`.

gmale:
> One issue with this would be I don't think you can rely on the sequence to be correct as you can see from this screenshot of my sync it's all over the place so they might get inserted in the wrong order (see `seqnum`).

FWIW: my machine behaves the same as @garethtdavies -- blocks have always streamed out-of-order during a reindex. As far as I understand, this shouldn't be a problem because there will be no duplicate blocks during that phase. In other words "old reorgs" have already self-corrected and "new reorgs" happen in real-time (so they should arrive in-sequence).

That being said, a slightly better fix might be to only do the replace only if the timestamp on the block is greater. For efficiency, we'd probably only want to check timestamps after an insert has failed and if the failed block is newer, do a replace, otherwise ignore, which basically boils down to not "relying on the sequence to be correct," as @garethtdavies suggested :)

PR #39, opened 2019-05-04
Correctly parse heights from script-encoded int64 values

### LINKED ISSUE #35: Issue ingesting blocks with block_height as -1

str4d:
This enables the parser to correctly extract heights for blocks 1 through 16.

Closes #35.


parser/block.go
```diff
@@ -62,21 +62,15 @@ func (b *block) GetHeight() int {
...
+	if heightNum < 0 {
 		return -1
 	}
 	// uint32 should last us a while (Nov 2018)
-	var blockHeight uint32
-	for i := heightLen - 1; i >= 0; i-- {
-		blockHeight <<= 8
-		blockHeight = blockHeight | uint32(heightBytes[i])
-	}
+	blockHeight := uint32(heightNum)
```
gtank: Since the read API returns a uint64, we need to do range checking on this conversion somewhere.

parser/internal/bytestring/bytestring.go
```diff
@@ -9,6 +9,11 @@ import (
 
 const MAX_COMPACT_SIZE uint64 = 0x02000000
 
+const OP_0 uint8 = 0x00
```
gtank: It would be better style to declare all of these in a block like:
```
const (
    OP_0 uint8 = 0x00
    ....
)```

Comment (gtank) [reply parser/internal/bytestring/bytestring.go]: Also, [idiomatically](https://golang.org/doc/effective_go.html#mixed-caps) these names should be camel case instead of capitals and underscores. They probably also don't need to be exported, so: `op0`, `op1Negate`, and so on. It may be an exception here if we're trying to match exact C++ code. I seem to have thought that previously, but really the above const _should_ have been called `maxCompactSize`.

File: parser/internal/bytestring/bytestring.go [hunk 3]
```diff
@@ -194,3 +199,42 @@ func (s *String) ReadUint64(out *uint64) bool {
...
+	firstByte := firstBytes[0]
+
+	var number uint64
+
+	if firstByte == OP_1NEGATE {
+		*num = -1
+		return true
+	} else if firstByte == OP_0 {
+		number = 0
+	} else if firstByte >= OP_1 && firstByte <= OP_16 {
```
gtank: It's not clear that this is correct. The C++ code seems to be directly comparing to the numbers 1 and 16, not these opcode values.

str4d: That's intentional. The referenced C++ code implements the serialization side; this function implements the parsing side. Parsing in `zcashd` would happen in the bowels of the script system.

str4d: I'll update the comments to clarify this, and reference a partial parser.

str4d:
Addressed @gtank's comments.

PR #40, opened 2019-05-04
Add PrevHash field to CompactBlocks

str4d:
This enables basic verification of chain validity for CompactBlocks without the full header.

This is an intentional backwards-incompatible change to the protobuf specification, and will require `lightwalletd` nodes to be regenerated from `zcashd`.


walletrpc/compact_formats.proto
```diff
@@ -13,9 +13,10 @@ message CompactBlock {
     uint32 protoVersion = 1; // the version of this wire format, for storage
     uint64 height = 2; // the height of this block
     bytes hash = 3;
-    uint32 time = 4;
-    bytes header = 5; // (hash and time) OR (full header)
-    repeated CompactTx vtx = 6; // compact transactions from this block
+    bytes prevHash = 4;
```
gtank: We're intentionally breaking this?

testdata/compact_blocks.json
```diff
@@ -3,36 +3,36 @@
...
-    "compact": "10b4d5111a204ff234f7b51971cbeb7719a1c32d1c7e1ed92afafed266a7b1ae235717df050020c59ce0dc05"
+    "compact": "10b4d5111a204ff234f7b51971cbeb7719a1c32d1c7e1ed92afafed266a7b1ae235717df050028c59ce0dc05"
```
gtank: What changed about these?

gtank:
Except for the trend of monotonically re-approaching the size of a normal block, LGTM :)

testdata/compact_blocks.json
gtank: Nevermind, the next commit replaces them anyway.

walletrpc/compact_formats.proto
str4d: Yes, because a) lightwalletd, like the rest of the stack, is in alpha and explicitly allowed to do this, and b) because it forces everyone to regenerate their lightwalletd servers, ensuring they all have `prevHash` fields.

testdata/compact_blocks.json
str4d: The protobuf field indicator changed due to the backwards-incompatible protobuf specification change. It altered only one or two bytes in the format here, while the next commit then inserted the `prevHash` fields.

walletrpc/compact_formats.proto
gtank: Just wanted it on the record. SGTM.

PR #43, opened 2019-06-07
Remove 0mq

### LINKED ISSUE #11: Write a zcash-rpc ingestor
This will let you spin up a light wallet without having to zmq the whole chain. It will also allow filling in gaps in a database, and generally provide some fallback options for everything. The implied json-rpc support will dovetail with #12.

Check the license situation, but the btcjson submodule of https://github.com/btcsuite/btcd looks promising.

mdr0id:
There exists a few issues with syncing properly on start and stop streams from a full zcashd node. This will now allow lightwalletd to gracefully start/stop with streams of blocks via `getblock` RPC calls. Additionally, this allows the lightwalletd to run from a full zcashd node without `-rescan` or power cycling zcashd. Ultimately, this completely removes 0mq logic.

The unsynced lightwalletd node will request blocks from current height in local db. If there is no existing database, the lightwalletd node will issue `getblock` from height 0.

Once lightwalletd has reached current height, it will poll every 60 seconds to check for the new block from zcashd. There exists a couple upcoming improvements that will be suited for mainnet and more stressful testnet cases, but for now this will remove a majority of existing issues reported in the field.

To test, please ensure your  `zcash.conf` specifies `rpcuser` and `rpcpassword`

(Important note: code assumes your full node is atleast at current height of local DB when you begin `getblock` routine ; I need to add logic for this case)

```
testnet=1
addnode=testnet.z.cash
rpcport = 18232
rpcbind = 127.0.0.1
rpcuser = username
rpcpassword = passw
```

On a side note, @str4d and I have looked into `getblock` performance improvements upstream and PRs are already in place.

Should close #11


mdr0id:
Kevin and I noticed an issue syncing from full nodes that are behind sapling height. I will add commits for this soon. However, I did add the logical commit to pull from Sapling height instead of 0 because the core code uses only Sapling. Pulling block prior to 280,000 on testnet doesn't make sense for this case.

cmd/ingest/main.go
```diff
@@ -90,107 +93,107 @@ func main() {
...
+			return nil, nil
 		}
+		return nil, errors.Wrap(rpcErr, "error requesting block")
+	}
 
-		switch string(topic) {
-
-		case opts.zmqTopic:
-			// there's an implicit mutex here
-			go handleBlock(db, sequence, body)
```
str4d: The goroutine disappears in this commit without any documentation or justification. Please remove the goroutine in its own commit (either before or after the 0MQ removal commit; I weakly prefer before).

cmd/ingest/main.go
```diff
@@ -1,18 +1,23 @@
...
+	"context"
 	"database/sql"
-	"encoding/binary"
 	"encoding/hex"
 	"flag"
 	"fmt"
 	"os"
+	"time"
+	"encoding/json"
+	"github.com/pkg/errors"
```
str4d: This should go into the second block of imports.

cmd/ingest/main.go
```diff
@@ -1,18 +1,23 @@
...
+	"strconv"
+	"strings"
```
str4d: Nit: the imports should remain sorted.

cmd/ingest/main.go
```diff
@@ -90,107 +93,107 @@ func main() {
...
-	// WARNING: The Socket is not thread safe. This means that you cannot
-	// access the same Socket from different goroutines without using something
-	// like a mutex.
-	sock, err := ctx.NewSocket(zmq.SUB)
-	if err != nil {
-		log.WithFields(logrus.Fields{
-			"error": err,
-		}).Fatal("couldn't create zmq context socket")
+		//Default to testnet, but user MUST specify rpcuser and rpcpassword in zcash.conf; no default
+		rpcClient, err = frontend.NewZRPCFromCreds("127.0.0.1:18232", " ", " ")
```
str4d: Shouldn't these be empty strings for name and password, not spaces?

cmd/ingest/main.go
```diff
@@ -90,107 +93,107 @@ func main() {
...
-		if len(msg) < 3 {
-			log.WithFields(logrus.Fields{
-				"msg": msg,
-			}).Warn("got unknown message type")
-			continue
+		if block != nil{
+			handleBlock(db , block)
+			height++
+			//TODO store block current/prev hash for formal reorg
+		}else{
```
str4d: Nit: whitespace around `else`.

cmd/ingest/main.go
```diff
@@ -90,107 +93,107 @@ func main() {
...
+				"error":  err,
+			}).Fatal("error with getblock")
 		}
-
-		if len(msg) < 3 {
-			log.WithFields(logrus.Fields{
-				"msg": msg,
-			}).Warn("got unknown message type")
-			continue
+		if block != nil{
```
str4d: Nit: whitespace after `nil`.

cmd/ingest/main.go
```diff
@@ -90,107 +93,107 @@ func main() {
...
+			handleBlock(db , block)
```
str4d: Nit: no whitespace in `db,`

cmd/ingest/main.go
```diff
@@ -90,107 +93,107 @@ func main() {
...
+	params[0] = json.RawMessage("\"" + strconv.Itoa(height) + "\"")
+	params[1] = json.RawMessage("0")
+	result, rpcErr := rpcClient.RawRequest("getblock", params)
+
+	var err error
+	var errCode int64
+
+	// For some reason, the error responses are not JSON
+	if rpcErr != nil {
+		errParts := strings.SplitN(rpcErr.Error(), ":", 2)
```
str4d: Add a comment saying why we are checking for error code `-8` (namely that it indicates we don't have the requested block height yet).

cmd/ingest/main.go
mdr0id: Put this in a new PR for now, see #46

cmd/ingest/main.go
mdr0id: Yes, I wasn't sure what to fudge there since the user has to specify these in `zcash.conf` explicitly. Please let me know if there is a default work around

str4d:
For anyone looking at using this PR to replace ZMQ, it is not ready for production. In particular, it does not handle reorgs yet (we need to manually implement functionality that we were previously relying on ZMQ to provide).

cmd/ingest/main.go
```diff
@@ -90,118 +93,156 @@ func main() {
...
-			"error": err,
-		}).Fatal("couldn't create zmq context socket")
+		//Default to testnet, but user MUST specify rpcuser and rpcpassword in zcash.conf; no default
+		rpcClient, err = frontend.NewZRPCFromCreds("127.0.0.1:18232", "", "")
+
+		if err != nil {
+			log.WithFields(logrus.Fields{
+				"error": err,
+			}).Fatal("couldn't start rpc connection")
+		}
```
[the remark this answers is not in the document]
mdr0id: Agreed the key error handling in the load NewRPCcreds or whatever, needs some TLC. I will address soon.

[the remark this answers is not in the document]
mdr0id: ```yes := cfg.Section("").HasKey("rpcbind")```

PR #53, opened 2019-08-30
add empty swift_prefix option value to .proto files

pacu:
Motivation.

proto-swift-gen script adds the package prefix on a snake case fashion to Swift Struct files. 
This makes the code difficult to read and has no convenience to us.

Example:
`CompactBlock` -> `Cash_Z_Wallet_SDK_RPC_CompactBlock`

When adding `option swift_prefix = "";` the resulting struct Name is `CompactBlock`


mdr0id:
utACK

PR #56, opened 2019-09-12
Reorg fixes

guardawallet:
There is PR #43 with our changes:
- handling reorg
- rpc params in start command (ingest, sever)
- zcashConfPath doesn't required
- logging reorg logic


cmd/ingest/main.go
```diff
@@ -22,20 +27,30 @@ var logger = logrus.New()
...
-	flag.StringVar(&opts.zmqTopic, "zmq-topic", "checkedblock", "the stream to listen to")
 	flag.StringVar(&opts.dbPath, "db-path", "", "the path to a sqlite database file")
 	flag.Uint64Var(&opts.logLevel, "log-level", uint64(logrus.InfoLevel), "log level (logrus 1-7)")
 	flag.StringVar(&opts.logPath, "log-file", "", "log file to write to")
+	flag.StringVar(&opts.zcashConfPath, "conf-file", "", "conf file to pull RPC creds from")
+	flag.Uint64Var(&opts.startHeight, "start-height", uint64(0), "start sync from this height (> 419200 for mainnet)")
+	flag.StringVar(&opts.rpcHost, "rpc-host", "", "RPC host of zcashd")
+	flag.StringVar(&opts.rpcPort, "rpc-port", "", "RPC port of zcashd")
+	flag.StringVar(&opts.rpcLogin, "rpc-login", "", "RPC login of zcashd")
+	flag.StringVar(&opts.rpcPass, "rpc-pass", "", "RPC password of zcashd")
```
gmale: this seems to be accepting the RPC password as a command-line option. If that's the case, then this password could be read from the process status info. It's probably better to leave these values in the `zcash.conf` file, since we need to access that file, anyway, and access to that file can be restricted via file permissions.

cmd/ingest/main.go
```diff
@@ -90,117 +105,158 @@ func main() {
...
-			"error":  err,
-			"stream": opts.zmqTopic,
-		}).Fatal("couldn't subscribe to stream")
 	}
 
-	connString := fmt.Sprintf("tcp://%s", opts.zmqAddr)
-
-	err = sock.Connect(connString)
-	if err != nil {
+	//ingest from Sapling testnet height
```
gmale: the comment here reads "testnet" but this is the "mainnet" start height. To prevent issues like this, in general, we may want to provide a more explicit way to handle `mainnet` v. `testnet` configuration. Perhaps, we would read that information from the `zcash.conf` file and choose a start height, accordingly. Rather than specifying it as a command-line option and then validating the input.

cmd/ingest/main.go
```diff
@@ -90,117 +105,158 @@ func main() {
...
+
+
+func handleBlock(db *sql.DB, block *parser.Block) {
+	nodeBlockHash := hex.EncodeToString(block.GetEncodableHash())
+
+	nodeBlockHeight := block.GetHeight()
+	localBlockHash, err := storage.GetBlockHash(db, nodeBlockHeight)
+
+	// if local block has different hash delete it before store new block
+	if localBlockHash != "" && localBlockHash != nodeBlockHash {
```
gmale: It appears that this, effectively, does an insert-replace for new blocks. That would fix single-block reorgs but it does not address larger ones. Given the change below to also [rewind the server by 4 blocks](https://github.com/zcash-hackworks/lightwalletd/pull/56/files#diff-0831e2c5d1d4855c0db87d71286f96a6R172) every time it reaches the chain tip, this would catch a lot of small reorgs but would fail on larger ones.

In other words, it seems these changes reduce the reorg problem but we could go a bit further and completely fix it (for instance by validating the prevHashes to confirm that every block is on-chain and the rewinding until we find the beginning of the chain split).

cmd/ingest/main.go
```diff
@@ -90,117 +105,158 @@ func main() {
...
-			continue
+		if block != nil {
+			handleBlock(db, block)
+			height++
+			if timeout_count > 0 {
+				timeout_count--
+			}
+			//TODO store block current/prev hash for formal reorg
+		} else {
+			height -= 4
```
gmale: Rewinding by 4 blocks every time we reach this point seems like it would cause every server to ingest over 4 times more data from zcashd because depending on the block time, this could rewind multiple times for each block. We may want to find a more efficient way to address the problem that this is trying to solve, instead.

cmd/ingest/main.go
```diff
@@ -90,117 +105,158 @@ func main() {
...
+				"height": height,
+				"error":  err,
+			}).Warn("error with getblock")
+			timeout_count++
+			if timeout_count == 3 {
+				log.WithFields(logrus.Fields{
+					"timeouts":  timeout_count,
+				}).Warn("unable to issue RPC call to zcashd node 3 times")
+				timeout_count = 0
+				time.Sleep(3 * 60 * time.Second)
```
gmale: By resetting the counter, this attempts to contact zcashd infinitely. We may want to handle this differently, perhaps exiting with an error, eventually.

cmd/ingest/main.go
```diff
@@ -1,18 +1,23 @@
...
 import (
+	"context"
 	"database/sql"
-	"encoding/binary"
 	"encoding/hex"
+	"encoding/json"
 	"flag"
 	"fmt"
 	"os"
+	"time"
```
gmale: Many of the changes in this Pull Request seem to originate from commits on the `remove_0mq` branch of in @mdr0id 's fork. So we may want to restructure this PR so that it builds ontop of that work so that it's easier to track the true history in the repository.

This would mean, @mdr0id would submit a PR to this repo and then, once that's merged, this PR can rebase on top of those changes.

gmale:
Thank you for submitting this PR! You’re tackling some real problems, however we also recently merged fixes for these issues, which are causing some conflicts. 

We encourage you to incorporate the latest changes from master whenever possible, specifically:  

- The latest reorg fixes. We handle them by checking prevHash values, which were recently added to CompactBlocks. Using the prevHash values is a more sound and efficient way that eliminates the need for repeatedly fetching the last couple blocks. 
- Username/password inputs for RPC. The current code uses the zcash.conf file rather than the command line for security reasons. 

After reviewing your PR, we’re considering: 1) adding support for mainnet and 2) making timeouts configurable so that the server can optionally exit if multiple errors are received while fetching blocks. We’ll reach out when these changes are added so the latest code on master  can be used directly.

PR #57, opened 2019-09-16
update broken gitlab yaml for cgo targets on alpine

mdr0id:
Update broken alpine image in CI


mdr0id:
Tested ACK: 
https://gitlab.com/mdr0id/lightwalletd/-/jobs/295720103

PR #62, opened 2019-09-18
Sanity check for why CI triggers for some folks and not others DO NOT MERGE

mdr0id:
Investigating annoying CI issues


mdr0id:
Used for sanity check. Closing

PR #81, opened 2019-09-25
pass artifacts to test stage and fix race test

mdr0id:
Resolve the broken job in test stage for Gitlab CI. Also pass the binaries for server and ingest to test stage.


mdr0id:
got test -race fail on Alpine is documented in several places: https://github.com/golang/go/issues/14481

Allow fail for now or will swap image type in test area if needed

PR #82, opened 2019-09-25
patch for race test in ci failing from no clang

mdr0id:


mdr0id:
Closing to archive. Issues found in this PR might be addressed formally now with existing Go profile tools

PR #165, opened 2020-01-08
Example usage of cobra and viper for configuration

benzcash:
This PR is designed to use the cobra and viper packages to provide additional configure options.

# Changes
- Added base main.go, so `go build` outputs `lightwalletd`
- Moved the server startup to the "root" command
- Create a subcommand for version
Example usage:
```
./lightwalletd version
lightwalletd version v0.2.0
```
- Tried to change the current flags as much as possible, but they are double-dashed now (we can add short options if desired)
```
Flags:
      --bind-addr string         the address to listen on (default "127.0.0.1:9067")
      --cache-size int           number of blocks to hold in the cache (default 80000)
      --config string            config file (default is current directory, lightwalletd.yaml)
  -h, --help                     help for lightwalletd
      --log-file string          log file to write to (default "./server.log")
      --log-level int            log level (logrus 1-7) (default 4)
      --no-tls-very-insecure     run without the required TLS certificate, only for debugging, DO NOT use in production
      --tls-cert string          the path to a TLS certificate (default "./cert.pem")
      --tls-key string           the path to a TLS key file (default "./cert.key")
      --zcash-conf-path string   conf file to pull RPC creds from (default "./zcash.conf")
```

# Added functionality
## Provide configuration from file
An example lightwalletd-example.yml is in the PR  
Usage:
```
./lightwalletd --config ./lightwalletd-example.yml
Using config file: /home/bwilson/go/src/github.com/zcash-hackworks/lightwalletd/lightwalletd.yml
```
Or, rename it to just `lightwalletd.yml` and it will be auto-discovered
```
$ ./lightwalletd 
Using config file: /home/bwilson/go/src/github.com/zcash-hackworks/lightwalletd/lightwalletd.yml
```

## Provide configuration from environment variables

The same options can be provided through environment variables
Example Usage:
```
TLS_CERT=/secrets/lightwallted/cert.pem \
TLS_KEY=/secrets/lightwallted/cert.key \
ZCASH_CONF_PATH=/srv/zcashd/zcash.conf \
CACHE_SIZE=10 \
LOG_FILE=/dev/stdout \
BIND_ADDR=0.0.0.0:9067 \
./lightwalletd
```
** NOTE ** We can preface these environmental variable with a string like `LIGHTWALLETD_` or `LWD_` to avoid collisions.


benzcash:
Moving to a local branch PR

Issue #215, opened 2020-03-18
Add docker image build and upload to CI


benzcash:
On new tags, and commits to the master branch, a docker image should be built, tagged and uploaded to Docker hub for deployment.

PR #227, opened 2020-03-24
Added tekton for Docker image build

benzcash:
Fixed the Dockerfile to use the correct binary file now that the build process changed.

Add tekton trigger and task to build a Docker image, tagged latest, from pushes to master.

Almost closes https://github.com/zcash/lightwalletd/issues/215. Trying to figure out the tagging for new releases.


tekton/triggerbinding.yml
```diff
@@ -0,0 +1,45 @@
...
+---
+apiVersion: tekton.dev/v1alpha1
+kind: TriggerTemplate
+metadata:
+  name: lightwalletd-dockerbuild-template
+spec:
+  params:
+  - name: newSHA
+    description: The git repository HEAD sha   
+  resourcetemplates:
```
charlieok: I'm curious why the `TaskRun` resource template defined here is identical to the `TaskRun` resource template in the `taskruns.yml` file above

benzcash: Because that way you can run the task independently from the trigger. It was really just part of the setup and test process, but I could see it being useful on it's own.

PR #246, opened 2020-04-26
Limit prometheus memory usage to 2G

charlieok:


mdr0id:
Not urgent. Just FYI for Ben when he returns as he was monitoring this from prior updates.

PR #289, opened 2020-06-17
Quick fix to darksidewalletd docs, will need improvements

defuse:


defuse:
Merging this without review because I urgently need it before a meeting with Aditya, I opened https://github.com/zcash/lightwalletd/issues/290 to come back later and make improvments.

PR #347, opened 2021-04-19
Verify rawtx is not NULL

adityapk00:
Verify rawtx before sending transaction


defuse:
We should merge this

LarryRuane (approved):
tested ACK

PR #394, opened 2022-05-30
eliminate redundancy, point URL to zingolabs

dannasessha:
I found some redundancy in the readme which I removed, and I have pointed the linking URL to our own `zecwalletlitelib` repo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,9 +18,9 @@ limitations of light wallets that use lightwalletd.
 
 # Overview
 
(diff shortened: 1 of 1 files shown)


LarryRuane (changes requested):
Thanks, this looks good, but could you add back the reference to https://github.com/adityapk00/zecwallet-lite-lib so we have both your new reference and that one? Perhaps something like:
```
[lightwalletd](https://github.com/zcash/lightwalletd) is a backend service that provides a bandwidth-efficient
interface to the Zcash blockchain for mobile and other wallets, such as
[Zingo Labs Zecwallet](https://github.com/zingolabs/zecwalletlitelib) or
[original Zecwallet](https://github.com/adityapk00/zecwallet-lite-lib).
```
Also, please rebase onto the latest `master` branch. If you'd like me to make these changes, just let me know (but it would be less work overall if you did them).

If you can make these changes, I'd be happy to merge this right away.

PR #475, opened 2024-02-24
Create CI.yaml in Github Actions

yasserisa:


yasserisa:
ACK

- When a PR is created, only the code is checked using some tests and it is verified that it can be compiled (including in a Docker image).

- If a release is made in the repository, the repository's tag is included in the label of the Docker image and the release is pushed to DockerHub.

CC: @LarryRuane

PR #487, opened 2024-05-28
Add optional proxy to exchange APIs for light clients

str4d:
This enables light clients to authenticate directly to exchange APIs without revealing their IP address.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -1,7 +1,9 @@
 package cmd
 
 import (
+	"context"
 	"fmt"
+	"io"
 	"net"
 	"net/http"
 	"os"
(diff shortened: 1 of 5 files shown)


cmd/root.go
```diff
@@ -423,3 +437,91 @@ func startHTTPServer(opts *common.Options) {
...
+			if ss.Target.ID == target.ID && ss.Target.Type == target.Type {
+				dialAddr = addr
+				break
+			}
+		}
+		if len(dialAddr) == 0 {
+			return fmt.Errorf("no matching dial port found for target: %s|%s", ss.Target.ID, ss.Target.Type)
+		}
+
+		// TODO: Enforce per-client (`ss.Addr`) rate limit.
```
str4d, line 482: We can probably use https://pkg.go.dev/golang.org/x/time/rate for this. What we want is:
- A request rate limiter per exchange, that we configure with their respective public API limits.
- A session rate limiter per client, that we configure with some low rate that is sufficient for the kinds of requests that honest light clients will be making ("infrequent" queries of market tickers).
  - We can probably initially assume that we have one session per request, but HTTP connection pooling on the light client side may break this assumption.
  - Once we have request-response detection, we can change this to (or add) a request rate limiter per client.
- Some mechanism for fairly distributing per-exchange request rate limit tokens amongst connected clients (subject to the per-client rate limiter).

str4d:
This PR will probably get closed, as instead of writing the client side of this (in the necessary Kotlin and Swift), I've integrated Arti into the Swift SDK for Tor support and am using that to fetch exchange rates (https://github.com/Electric-Coin-Company/zcash-light-client-ffi/pull/142, https://github.com/Electric-Coin-Company/zcash-swift-wallet-sdk/pull/1446).

PR #499, opened 2024-09-12
Update Dockerfile

y4ssi:
The purpose of this change is to simplify the Dockerfile, enabling the Docker image to be run with a custom user without the need to build a custom image, but rather using the official version of the public image


daira (approved):
utACK

PR #523, opened 2025-06-05
refactor: use a more modern writing style

rustrover:
Optimize code using a more modern writing style.   Official support from Go, for more details visit https://pkg.go.dev/golang.org/x/tools/gopls/internal/analysis/modernize.

### DIFF
--- a/common/darkside.go
+++ b/common/darkside.go
@@ -9,6 +9,7 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"slices"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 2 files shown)


rustrover:
> This is a nice improvement, thanks! I noted one correction that is needed. I ran the tool myself (wasn't aware of it, thanks for that too), and it finds many instances of:
> 
> ```
> for loop can be modernized using range over int
> ```
> 
> and a few instances of:
> 
> ```
> interface{} can be replaced by any
> ```
> 
> I haven't investigated what these are about, but is it worth fixing those too?

Thank you for your interest in this. I have disabled several checks, such as the rangeint and efaceany. The  `rangeint`  check optimizes code like this:

```go
for i := 0; i < 6; i++ { }
```

to 

`for range 6 {}`. 



The `efaceany`  check changes `interface{}` to `any`. 

I don't think either of these changes is very necessary (in fact, the official Go tools added the disable feature specifically for cases like `efaceany`, where the modifications are extensive but not very meaningful).

LarryRuane (approved):
ACK a493f7aef78005b2bdf29cd54d516a32420b3b93

PR #524, opened 2025-06-23
chore: fix some function names in comment

curlwget:
fix some function names in comment

### DIFF
--- a/common/darkside.go
+++ b/common/darkside.go
@@ -167,7 +167,7 @@ func DarksideReset(sa int, bi, cn string, sst, sot uint32) error {
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


LarryRuane:
Thanks — both comments were stale, and #526 reported the `GetTaddressTransactions` one as well.

This PR conflicts with master, so rather than ask for a rebase I've fixed both in one commit and credited both reports. The `GetTaddressTransactions` comment also needed its trailing "NB, this method is misnamed" line dropped: that note described `GetTaddressTxids`, which still exists as a deprecated wrapper further down and carries the explanation there.

Closing in favor of that commit.

PR #526, opened 2025-07-06
chore: make function comment match function name

rustfix:
make function comment match function name


LarryRuane:
Thanks — the comment was indeed stale, and so was the one in `common/darkside.go` that #524 caught.

This PR conflicts with master, so rather than ask for a rebase I've fixed both in one commit and credited both reports. The `GetTaddressTransactions` comment also needed its trailing "NB, this method is misnamed" line dropped: that note described `GetTaddressTxids`, which still exists as a deprecated wrapper further down and carries the explanation there.

Closing in favor of that commit.

PR #553, opened 2026-03-10
Fix infinite loop in darkside setPrevhash (#552)

### LINKED ISSUE #552: Darkside setPrevhash causes infinite add/reorg loop in BlockIngestor
**What is the bug?**
When running the darksidewalletd tutorial reorg example, calling ApplyStaged causes the block ingestor to enter an infinite loop, repeatedly adding and dropping the same block:

  {"app":"lightwalletd","level":"info","msg":"Adding block to cache 663151 17c8c93591658f5ffc8e30f138cadcdf66c4ca4b60604427984d023449e71da2"}
  {"app":"lightwalletd","level":"info","msg":"REORG: dropping block 663151 17c8c93591658f5ffc8e30f138cadcdf66c4ca4b60604427984d023449e71da2"}
  {"app":"lightwalletd","level":"info","msg":"Adding block to cache 663151 17c8c93591658f5ffc8e30f138cadcdf66c4ca4b60604427984d023449e71da2"}
  {"app":"lightwalletd","level":"info","msg":"REORG: dropping block 663151 17c8c93591658f5ffc8e30f138cadcdf66c4ca4b60604427984d023449e71da2"}

  This is reproducible by following the tutorial in docs/darksidewalletd.md and executing the ApplyStaged step at height 663210.

**Additional context**
The bug is in setPrevhash() in common/darkside.go. The function parses each block from its raw bytes before writing the corrected prevhash into those bytes. It then computes the block hash from the parsed struct, which still contains the old (incorrect) prevhash. This means subsequent blocks receive an incorrect prevhash that doesn't match the actual hash of the preceding block.

When BlockIngestor later fetches blocks via getBlockFromRPC, it re-parses from the (now corrected) raw bytes and computes the true hash. The next block's prevhash (set using the wrong hash) doesn't match, triggering a reorg. The reorg drops the block, which is then re-added, creating an infinite loop.

**Solution**
Move the copy(activeBlock.bytes[4:4+32], prevhash[:]) call to execute before block.ParseFromSlice(activeBlock.bytes), so the block is parsed from the updated bytes and GetEncodableHash() computes the correct hash including the new prevhash.

mnm458:
### Description:
Fix infinite loop in darkside setPrevhash by computing block hash after updating prevhash bytes.

### Summary
`setPrevhash()` was parsing each block's raw bytes *before* writing the corrected prevhash, then computing the block hash from the stale parsed struct. This caused subsequent blocks to have incorrect prevhash values, leading `BlockIngestor` into an infinite add/reorg loop when `ApplyStaged` was called.

The fix moves the `copy()` of the prevhash to occur before `ParseFromSlice()`, so the hash is computed from the updated bytes.

Closes #552

### Documentation
No external documentation changes required. The fix corrects internal behavior to match the existing documented darkside tutorial in `docs/darksidewalletd.md`.

### Test Plan
  - [x] Added regression test `TestSetPrevhashChainConsistency` in `common/darkside_test.go`
  - [x] Test verifies that each block's prevhash matches the preceding block's hash after `setPrevhash` runs
  - [x] Test fails against the old (buggy) code and passes with the fix
  - [x] All existing tests pass (`go test ./...`)
  - [ ] Manual verification: run the darkside tutorial from `docs/darksidewalletd.md` and confirm `ApplyStaged` no longer produces infinite log output

### DIFF
--- a/common/darkside.go
+++ b/common/darkside.go
@@ -232,6 +232,11 @@ func setPrevhash() {
 	prevhash := hash32.Nil
 	for _, activeBlock := range state.activeBlocks {
 		// Set this block's prevhash.
+		if prevhash != hash32.Nil {
+			copy(activeBlock.bytes[4:4+32], prevhash[:])
+		}
+		// Re-parse the block from the (possibly modified) bytes
+		// so that the hash we compute includes the updated prevhash.
 		block := parser.NewBlock()
 		rest, err := block.ParseFromSlice(activeBlock.bytes)
 		if err != nil {
@@ -239,9 +244,6 @@ func setPrevhash() {
 		}
 		if len(rest) != 0 {
 			Log.Fatal(errors.New("block is too long"))
-		}
-		if prevhash != hash32.Nil {
-			copy(activeBlock.bytes[4:4+32], prevhash[:])
 		}
 		prevhash = block.GetEncodableHash()
 		Log.Info("Darkside active block height ", block.GetHeight(), " hash ",
--- a/common/darkside.go
+++ b/common/darkside.go
@@ -235,8 +235,6 @@ func setPrevhash() {
 		if prevhash != hash32.Nil {
 			copy(activeBlock.bytes[4:4+32], prevhash[:])
 		}
-		// Re-parse the block from the (possibly modified) bytes
-		// so that the hash we compute includes the updated prevhash.
 		block := parser.NewBlock()
 		rest, err := block.ParseFromSlice(activeBlock.bytes)
 		if err != nil {
--- a/common/darkside_test.go
+++ b/common/darkside_test.go
@@ -1,0 +1,83 @@
+package common
+
+import (
+	"os"
+	"testing"
+
+	"github.com/zcash/lightwalletd/hash32"
+	"github.com/zcash/lightwalletd/parser"
+)
+
+// TestSetPrevhashChainConsistency verifies that after setPrevhash runs,
+// each block's prevhash matches the actual hash of the preceding block.
+// This is a regression test for issue #552, where setPrevhash computed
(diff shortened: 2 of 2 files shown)


common/darkside.go
```diff
@@ -232,6 +232,11 @@ func setPrevhash() {
 	prevhash := hash32.Nil
 	for _, activeBlock := range state.activeBlocks {
 		// Set this block's prevhash.
+		if prevhash != hash32.Nil {
+			copy(activeBlock.bytes[4:4+32], prevhash[:])
+		}
+		// Re-parse the block from the (possibly modified) bytes
```
nullcopy: Unless I missed it, this is the first parse, not a re-parse. IMO, just drop this comment entirely, the code is self explanatory here

nullcopy (dismissed):
utACK c7886be69eaf8b26a5d2467a0fc294d9cb5b743f modulo the confusing comment

common/darkside.go
mnm458: Good catch, removed. The intent was to clarify why the parse must happen after the `prevhash` write, but the comment was doing too much. The code speaks for itself here

nullcopy (approved):
utACK 692f81bea2e594c3a95a2d6888f4a17868e96c3a

nullcopy:
Apparently the build check got changed. Need to rebase on main for this check to run :facepalm:

mnm458:
> Apparently the build check got changed. Need to rebase on main for this check to run 🤦

Done, rebased on master! I think you might need to help approve and run the check

mnm458:
@nullcopy Rebased on latest master again. Could you please help approve so we can merge?

nullcopy (approved):
utACK d95a5ccd11b4d2f65294f00f7512fa5b940c4e76

PR #560, opened 2026-05-03
fix(frontend): cancel GetBlockRange producer goroutine on client disconnect

### LINKED ISSUE #559: GetBlockRange/GetBlockRangeNullifiers leak producer goroutine on client disconnect
### Summary

`GetBlockRange` (`frontend/service.go:230`) and `GetBlockRangeNullifiers` (`frontend/service.go:256`) spawn a producer goroutine that sends blocks on an **unbuffered** channel without context-cancellation hygiene. When a gRPC client cancels mid-stream (TCP RST, deadline, peer crash), the handler returns but the producer goroutine blocks forever on `blockOut <- block` because no consumer remains. Every cancelled stream leaks one goroutine indefinitely, retaining a `*BlockCache` reference and one in-flight `*CompactBlock` until process exit.

A PR with the fix and a sibling-handler audit is incoming and will reference this issue.

### Vulnerable code

`frontend/service.go:230-252` (`GetBlockRange`):

```go
func (s *lwdStreamer) GetBlockRange(span *walletrpc.BlockRange, resp ...) error {
    blockChan := make(chan *walletrpc.CompactBlock)   // UNBUFFERED
    if span.Start == nil || span.End == nil { ... }
    errChan := make(chan error)                        // UNBUFFERED
    go common.GetBlockRange(s.cache, blockChan, errChan, span)

    for {
        select {
        case err := <-errChan:
            return err
        case cBlock := <-blockChan:
            err := resp.Send(cBlock)
            if err != nil { return err }                // <-- handler returns
        }
        // No <-resp.Context().Done() case.
    }
}
```

Producer at `common/common.go:572-604`:

```go
func GetBlockRange(cache *BlockCache, blockOut chan<- *walletrpc.CompactBlock,
                   errOut chan<- error, span *walletrpc.BlockRange) {
    // No ctx parameter at all.
    ...
    for i := low; i <= high; i++ {
        ...
        block, err := GetBlock(cache, j)
        if err != nil { errOut <- err; return }   // BLOCKING send, no listener
        ...
        blockOut <- block                          // BLOCKING send, no listener
    }
    errOut <- nil                                  // BLOCKING send, no listener
}
```

After `resp.Send(cBlock)` returns an error (client cancelled), the next blocking send in the producer parks forever.

### Sibling-handler audit (clean)

I scanned `frontend/service.go` for the same `make(chan ...)` + `go ...` antipattern. Only `GetBlockRange` and `GetBlockRangeNullifiers` exhibit it. Other streaming handlers (`GetTaddressTransactions`, `GetTaddressBalanceStream`, `GetMempoolStream`, `GetMempoolTx`, `GetSubtreeRoots`, `GetAddressUtxosStream`) iterate inline against a callback or synchronous loop and propagate cancellation through `resp.Send` errors.

### Reproducer

Minimum Go reproducer (~30 lines):

```go
package main

import (
    "context"
    "fmt"
    "time"

    "github.com/zcash/lightwalletd/walletrpc"
    "google.golang.org/grpc"
    "google.golang.org/grpc/credentials/insecure"
)

func main() {
    conn, _ := grpc.Dial("127.0.0.1:9067",
        grpc.WithTransportCredentials(insecure.NewCredentials()))
    defer conn.Close()
    client := walletrpc.NewCompactTxStreamerClient(conn)

    for i := 0; i < 10000; i++ {
        ctx, cancel := context.WithCancel(context.Background())
        stream, err := client.GetBlockRange(ctx, &walletrpc.BlockRange{
            Start: &walletrpc.BlockID{Height: 1},
            End:   &walletrpc.BlockID{Height: 3000000},
        })
        if err != nil { continue }
        _, _ = stream.Recv()  // receive one block
        cancel()              // then cancel
        if i%100 == 0 { fmt.Printf("iter %d\n", i) }
        time.Sleep(10 * time.Millisecond)
    }
}
```

Verify the leak via pprof (lightwalletd exposes `:9068/debug/pprof/`):

```
curl 'http://127.0.0.1:9068/debug/pprof/goroutine?debug=1' \
  | grep -A2 "common.GetBlockRange" | head
```

The number of goroutines parked at `common.GetBlockRange` will equal the number of cancelled streams.

### Suggested fix

Pass `resp.Context()` to the producer; have both sides respect cancellation via `select { case ch <- v: case <-ctx.Done(): }`. The producer's first iteration after the consumer leaves then exits cleanly instead of parking. Patch attached on the incoming PR.

### Per-leak memory cost

Each leaked goroutine retains, at minimum:

- One Go goroutine stack: 8 KiB initial (grows up to 1 GiB by Go runtime defaults but typically stays at 8-32 KiB for this code path).
- One `*walletrpc.CompactBlock` value parked on the channel send: typical sapling/orchard-era block compact is ~10-200 KiB depending on tx density (peak heights observed at >500 KiB).
- One `*BlockCache` reference (shared, no per-leak cost beyond the pointer itself).
- The captured loop-local `i`, `j`, `block`, `err`, `errMsg` variables: <1 KiB.

Per-leak memory: roughly **20-220 KiB** at typical heights, peaking at ~530 KiB.

For a sustained 1-rps cancellation attack over 1 hour: ~3,600 leaked goroutines × ~100 KiB mean ≈ **360 MiB** of unreclaimable RSS, plus runtime overhead.

This is consistent with goroutine-leak DoS observations in similar gRPC streaming services (cf. grpc-go issue #2379 and follow-ups). Exact numbers depend on chain height and tx density and would need to be confirmed by running the reproducer below against a real lightwalletd instance.

### Severity

Medium DoS. Single-shot per stream; sustained attack required to push RSS into operational concern (a public-bound deployment under steady cancellation load reaches 100s of MiB unreclaimable within hours). lightwalletd defaults to `127.0.0.1:9067` (loopback) — production deployments fronted by a public load-balancer or running on `0.0.0.0` are affected; loopback-only deployments require local code execution to reach.

### AI disclosure

This finding was identified, the patch was drafted, and the issue body was written with assistance from Claude (Anthropic). I reviewed each step and am the responsible author.

dingledropper:
## Summary

Fixes #559.

`GetBlockRange` and `GetBlockRangeNullifiers` (`frontend/service.go:230, 256`) spawn a producer goroutine that sends blocks on an unbuffered channel without context-cancellation hygiene. When a gRPC client cancels mid-stream, the handler returns but the producer blocks forever on the next `blockOut <- block`, leaking one goroutine per cancelled stream and retaining a `*BlockCache` reference + one `*CompactBlock` until process exit.

## Fix

- `common/common.go`: add `ctx context.Context` as first parameter; wrap each blocking send in `select { case ch <- v: case <-ctx.Done(): return }` so the producer exits cleanly when the consumer leaves.
- `frontend/service.go` (both handlers): pass `resp.Context()` to the producer and add a `<-ctx.Done()` case to the consumer's select loop.

Sibling-handler audit: scanned `frontend/service.go` for the same `make(chan ...)` + `go ...` antipattern. Only the two named handlers exhibit it; other streaming RPCs (`GetTaddressTransactions`, `GetTaddressBalanceStream`, `GetMempoolStream`, `GetMempoolTx`, `GetSubtreeRoots`, `GetAddressUtxosStream`) iterate inline and propagate cancellation through `resp.Send` errors.

## Test plan

- [ ] `go build ./...` (I do not have a Go toolchain locally; please verify)
- [ ] `go test ./...`
- [ ] Reproduce the leak by running the PoC in #559 against an unmodified build (pprof shows N goroutines parked at `common.GetBlockRange`)
- [ ] Re-run the PoC against this branch (pprof shows 0 leaked goroutines after cancellation)
- [ ] Existing `GetBlockRange` integration tests pass

## Notes

- Patch passes `ctx` only to the producer's send-paths. Out-of-scope follow-up: thread `ctx` through `GetBlock` / the upstream zcashd JSON-RPC client so an in-flight upstream call cancels with the gRPC stream. Happy to follow up if you'd like.
- The third blocking send (`errOut <- nil` on success at the end of the loop) was also wrapped in `select` for symmetry, even though that path can only be reached when the consumer is still alive.

## AI disclosure

This finding was identified, the patch was drafted, and the PR description was written with assistance from Claude (Anthropic). I reviewed each step and am the responsible author.

### DIFF
--- a/frontend/service.go
+++ b/frontend/service.go
@@ -229,16 +229,21 @@ func (s *lwdStreamer) GetBlockNullifiers(ctx context.Context, id *walletrpc.Bloc
 // 'end' inclusively.
 func (s *lwdStreamer) GetBlockRange(span *walletrpc.BlockRange, resp walletrpc.CompactTxStreamer_GetBlockRangeServer) error {
 	common.Log.Debugf("gRPC GetBlockRange(%+v)\n", span)
-	blockChan := make(chan *walletrpc.CompactBlock)
 	if span.Start == nil || span.End == nil {
 		return status.Error(codes.InvalidArgument,
 			"GetBlockRange: must specify start and end heights")
 	}
+	ctx := resp.Context()
+	blockChan := make(chan *walletrpc.CompactBlock)
 	errChan := make(chan error)
-	go common.GetBlockRange(s.cache, blockChan, errChan, span)
+	go common.GetBlockRange(ctx, s.cache, blockChan, errChan, span)
 
 	for {
 		select {
+		case <-ctx.Done():
+			// Client cancelled / deadline exceeded; the producer's select-on-ctx
+			// will unblock its in-flight send and exit.
+			return ctx.Err()
 		case err := <-errChan:
 			// this will also catch context.DeadlineExceeded from the timeout
 			return err
@@ -255,7 +260,6 @@ func (s *lwdStreamer) GetBlockRange(span *walletrpc.BlockRange, resp walletrpc.C
 // the actions contain only nullifiers (a subset of the full compact block).
 func (s *lwdStreamer) GetBlockRangeNullifiers(span *walletrpc.BlockRange, resp walletrpc.CompactTxStreamer_GetBlockRangeNullifiersServer) error {
 	common.Log.Debugf("gRPC GetBlockRangeNullifiers(%+v)\n", span)
-	blockChan := make(chan *walletrpc.CompactBlock)
 	if span.Start == nil || span.End == nil {
 		return status.Error(codes.InvalidArgument,
 			"GetBlockRangeNullifiers: must specify start and end heights")
@@ -269,11 +273,17 @@ func (s *lwdStreamer) GetBlockRangeNullifiers(span *walletrpc.BlockRange, resp w
 		}
 	}
 	span.PoolTypes = filtered
+	ctx := resp.Context()
+	blockChan := make(chan *walletrpc.CompactBlock)
 	errChan := make(chan error)
-	go common.GetBlockRange(s.cache, blockChan, errChan, span)
+	go common.GetBlockRange(ctx, s.cache, blockChan, errChan, span)
 
 	for {
 		select {
+		case <-ctx.Done():
+			// Client cancelled / deadline exceeded; the producer's select-on-ctx
+			// will unblock its in-flight send and exit.
+			return ctx.Err()
 		case err := <-errChan:
 			// this will also catch context.DeadlineExceeded from the timeout
 			return err
--- a/common/common.go
+++ b/common/common.go
@@ -5,6 +5,7 @@
 package common
 
 import (
+	"context"
 	"encoding/hex"
 	"encoding/json"
 	"errors"
@@ -570,9 +571,17 @@ func filterBlockPool(vtx []*walletrpc.CompactTx, poolTypes []walletrpc.PoolType)
 }
 
 // GetBlockRange returns a sequence of consecutive blocks in the given range.
-func GetBlockRange(cache *BlockCache, blockOut chan<- *walletrpc.CompactBlock, errOut chan<- error, span *walletrpc.BlockRange) {
+//
+// The `ctx` parameter is used to abort iteration when the gRPC client cancels
+// the stream. Without it, the producer goroutine would block indefinitely on
+// the unbuffered `blockOut` send after the consumer (the gRPC handler) returns,
+// leaking one goroutine per cancelled stream.
+func GetBlockRange(ctx context.Context, cache *BlockCache, blockOut chan<- *walletrpc.CompactBlock, errOut chan<- error, span *walletrpc.BlockRange) {
 	if slices.Contains(span.PoolTypes, walletrpc.PoolType_POOL_TYPE_INVALID) {
-		errOut <- fmt.Errorf("GetBlockRange: invalid pool type requested")
+		select {
+		case errOut <- fmt.Errorf("GetBlockRange: invalid pool type requested"):
+		case <-ctx.Done():
+		}
 		return
 	}
 	// Go over [start, end] inclusive
@@ -591,16 +600,26 @@ func GetBlockRange(cache *BlockCache, blockOut chan<- *walletrpc.CompactBlock, e
 
 		block, err := GetBlock(cache, j)
 		if err != nil {
-			errOut <- err
+			select {
+			case errOut <- err:
+			case <-ctx.Done():
+			}
 			return
 		}
 		block.Vtx = filterBlockPool(block.Vtx, span.PoolTypes)
 
 		// Note that we do want to return blocks that have had all of its transactions filtered,
 		// as we have done in the past.
-		blockOut <- block
+		select {
+		case blockOut <- block:
+		case <-ctx.Done():
+			return
+		}
 	}
-	errOut <- nil
+	select {
+	case errOut <- nil:
+	case <-ctx.Done():
+	}
 }
 
 // ParseRawTransaction converts between the JSON result of a `zcashd`
--- a/common/common_test.go
+++ b/common/common_test.go
@@ -6,6 +6,7 @@ package common
 import (
 	"bufio"
 	"bytes"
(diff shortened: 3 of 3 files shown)


frontend/service.go
```diff
@@ -229,16 +229,21 @@ func (s *lwdStreamer) GetBlockNullifiers(ctx context.Context, id *walletrpc.Bloc
 // 'end' inclusively.
 func (s *lwdStreamer) GetBlockRange(span *walletrpc.BlockRange, resp walletrpc.CompactTxStreamer_GetBlockRangeServer) error {
 	common.Log.Debugf("gRPC GetBlockRange(%+v)\n", span)
-	blockChan := make(chan *walletrpc.CompactBlock)
 	if span.Start == nil || span.End == nil {
 		return status.Error(codes.InvalidArgument,
 			"GetBlockRange: must specify start and end heights")
 	}
+	ctx := resp.Context()
+	blockChan := make(chan *walletrpc.CompactBlock)
```
LarryRuane, line 237: Nice that you're moving this down to where it's used (before the unrelated error return)

LarryRuane (changes requested):
This fix works perfectly. I verified it by running the reproduce program given in #559 and then pausing `lightwalletd` in the debugger, and observed that there are a huge number of those blocked threads in `GetBlockRange`: 
Image

The unit tests that call `GetBlockRange` no longer compile, due to the missing first argument, so we can add this to your PR (it should go into the same commit):

```diff
 --- a/common/common_test.go
+++ b/common/common_test.go
@@ -6,6 +6,7 @@ package common
 import (
        "bufio"
        "bytes"
+       "context"
        "encoding/json"
        "errors"
        "fmt"
@@ -454,7 +455,7 @@ func TestGetBlockRange(t *testing.T) {
                Start: &walletrpc.BlockID{Height: 380640},
                End:   &walletrpc.BlockID{Height: 380642},
        }
-       go GetBlockRange(testcache, blockChan, errChan, blockRange)
+       go GetBlockRange(context.Background(), testcache, blockChan, errChan, blockRange)
 
        // read in block 380640
        select {
@@ -557,7 +558,7 @@ func TestGetBlockRangeReverse(t *testing.T) {
                Start: &walletrpc.BlockID{Height: 380642},
                End:   &walletrpc.BlockID{Height: 380640},
        }
-       go GetBlockRange(testcache, blockChan, errChan, blockRange)
+       go GetBlockRange(context.Background(), testcache, blockChan, errChan, blockRange)
 
        // read in block 380642
        select {
```
That's the simplest way to fix the unit test, or we can modify these tests to actually test recovery from cancellation. All tests need to compile and run successfully before merging this PR (so I'm requesting that this be done).

In the interest of time, we could do the simple fix here in this PR, and improve the test coverage in a follow-up PR. Or the testing can be improved in this PR. I believe you've written a new test, `range_stream_leak_poc_test.go`, that we can add to this PR, but I haven't reviewed it.

dingledropper:
Thanks for verifying the fix in the debugger @LarryRuane — and apologies for the unit-test compile breakage, that should have been caught locally before opening.

Force-pushed an amended commit (`0b155d5`) per your "should go into the same commit" note. The amend applies your suggested compile-fix diff verbatim to `common/common_test.go`: adds the `context` import and passes `context.Background()` as the first arg to `GetBlockRange` in both `TestGetBlockRange` and `TestGetBlockRangeReverse`. Both tests pass locally.

Happy to put together a `range_stream_leak_poc_test.go` regression test in a follow-up PR if you'd like the test coverage. Re-requesting review.

Thanks much, dingledropper

LarryRuane (approved):
tACK 0b155d5a2c8b7171d63e956c2a6ba09ce84587e6

Thanks for this PR! All tests are now passing. I think, as a follow-up PR for the regression test, of course, there's no harm in doing that. But I'm not sure it's worth the effort, so I'll leave that to you. The reason I'm not sure is that regression tests are great for covering things that might be broken by unrelated changes, but I think that's unlikely in this case.

dingledropper:
Thanks @LarryRuane — and good call on the regression test; agreed the unrelated-change risk is low enough here that focused commit history wins over coverage-for-coverage's-sake. Will leave it.

Appreciate the quick turnaround. Thanks much, dingledropper

PR #562, opened 2026-05-29
Bound parser counts before allocation

Michae2xl:
## Summary
- reject CompactSize counts or lengths that cannot fit in the remaining parser input before allocating slices
- cover block transaction counts, block header Equihash solution length, transparent inputs and outputs, Sapling v4/v5 spends and outputs, JoinSplits, and Orchard actions
- preserve valid parsing behavior while returning earlier errors for malformed or truncated input

## Testing
- `GOCACHE=/tmp/zcash-lightwalletd-go-build go test ./parser`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,14 @@ The most recent changes are listed first.
 
 ### Fixed
 
+- Bound element counts in the block and transaction parser against the
+  remaining input before allocating. A malformed or truncated serialization
+  can declare far more elements (transaction inputs and outputs, Sapling
+  spends and outputs, Orchard actions, JoinSplits, and the block's
+  transaction count) than the input could possibly contain; sizing a slice
+  from such a count previously allocated gigabytes before the first element
+  failed to parse, exhausting memory (GHSA-ph5v-77v6-j498).
+
 - Make `common.RawRequest` context-aware so cancelled `GetBlockRange` /
   `GetBlockRangeNullifiers` streams abort in-flight zcashd JSON-RPC calls
   instead of holding a goroutine and RPC connection until the request
--- a/parser/block.go
+++ b/parser/block.go
@@ -72,6 +72,7 @@ func (b *Block) GetDisplayPrevHashString() string {
 
 // see https://github.com/zcash/lightwalletd/issues/17#issuecomment-467110828
 const genesisTargetDifficulty = 520617983
+const minTransactionWireBytes = 5 // 4-byte header + at least one CompactSize byte
 
 // GetHeight extracts the block height from the coinbase transaction. See
 // BIP34. Returns block height on success, or -1 on error.
@@ -138,6 +139,9 @@ func (b *Block) ParseFromSlice(data []byte) (rest []byte, err error) {
 	var txCount int
 	if !s.ReadCompactSize(&txCount) {
 		return nil, errors.New("could not read tx_count")
+	}
+	if err := rejectCountExceedingRemaining("tx_count", txCount, len(s), minTransactionWireBytes); err != nil {
+		return nil, err
 	}
 	data = []byte(s)
 
--- a/parser/block.go
+++ b/parser/block.go
@@ -72,7 +72,6 @@ func (b *Block) GetDisplayPrevHashString() string {
 
 // see https://github.com/zcash/lightwalletd/issues/17#issuecomment-467110828
 const genesisTargetDifficulty = 520617983
-const minTransactionWireBytes = 5 // 4-byte header + at least one CompactSize byte
 
 // GetHeight extracts the block height from the coinbase transaction. See
 // BIP34. Returns block height on success, or -1 on error.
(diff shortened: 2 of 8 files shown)


LarryRuane:
Thanks for this PR, but I don't think these changes are necessary, since we already completely trust our local full node (`zcashd` or `zebrad`). These changes protect us from a malicious or buggy full node, but that's not something we need protection from. If the changes in this PR are needed, I believe we would need to make many other similar hardening improvements to be consistent, and I don't think those are worth doing; that's just not the model we're following.

Note that this is very different from hardening against our gRPC clients; those are beneficial because we don't trust those clients. But we do trust our local full node.

Let me get the opinions of @nuttycom and @daira.

LarryRuane:
Thanks for this — the approach is right and the arithmetic is careful. I verified every constant against the corresponding parser and they're all exactly correct:

| Constant | Fields | |
|---|---|---|
| `minTxInWireBytes` 41 | 32 hash + 4 index + ≥1 script len + 4 sequence | ✔ |
| `minTxOutWireBytes` 9 | 8 value + ≥1 script len | ✔ |
| `minSaplingV4SpendBytes` 384 | 32+32+32+32+192+64 | ✔ |
| `minSaplingV4OutputBytes` 948 | 32+32+32+580+80+192 | ✔ |
| `minSaplingV5SpendBytes` 96 | 32+32+32 | ✔ |
| `minSaplingV5OutputBytes` 756 | 32+32+32+580+80 | ✔ |
| `minOrchardActionBytes` 820 | 32+32+32+32+32+580+80 | ✔ |
| `minJoinSplit{Groth,PHGR}Bytes` 1698/1802 | 8+8+32+64+64+32+32+64+{192,296}+1202 | ✔ |

Using `count > remaining/minElementBytes` is the right shape: the truncating division only ever loosens the bound, so it can't reject input that would have parsed.

The part that actually matters is `ParseTransparent` and `parsePreV5`, which had no cap at all — `make([]txIn, txInCount)` at `maxCompactSize` is ~2.1 GB. `parseV5` was already covered by its `>= (1 << 16)` checks.

Worth stating in the PR description: every caller of these parsers takes its input from a trusted local zcashd (`common/common.go` via `getblock`, `frontend/service.go` via `getrawtransaction`) or from darkside test mode. So this is defense-in-depth rather than a remotely-triggerable bug, which is a fine thing to merge but shouldn't be read as urgent.

## Rebased and pushed

The branch had gone stale against #563's ZIP 229 v6 work, which moved the v5 Sapling and Orchard parsing out of `parseV5` into the shared `parseSaplingBundle` and `parseOrchardActionShapeBundle`. I've rebased and pushed, keeping your commit intact with a follow-up on top.

Pleasant side effect: the checks now live in the shared helpers, so they cover v6 and the Ironwood bundle too.

The rebase surfaced two things:

- **Compile error.** `parseOrchardActionShapeBundle` returns three values, but the relocated check did `return nil, err`. Fixed in the rebased commit so it builds standalone.
- **Wrong pool name.** That same check hardcoded `"nActionsOrchard"`, but ZIP 229 shares the function between Orchard and Ironwood and every neighbouring error uses the `pool` parameter. Now `"nActions"+pool`, with a test over both.

## Follow-up commit

- Moved the constants and helper into a new `parser/bounds.go`. They're used from `block.go` and `block_header.go` too, and `minTransactionWireBytes` was declared off on its own in `block.go`.
- Reworded the error to report what's required rather than what remains. The old form read as a contradiction whenever the count was smaller than the bytes left: `tx_in_count 2 exceeds remaining input length 45`. Now: `tx_in_count 2 requires at least 82 bytes, but only 45 remain`.
- Widened `count*minElementBytes` in the message to `int64` — at `maxCompactSize` (2²⁵) × 1802 that overflows a 32-bit `int`, and `ReadCompactSize` explicitly contemplates 32-bit builds.
- Guarded the helper against a non-positive element size, which would panic on the division.
- `minTransactionWireBytes` 5 → 10, the true minimum (v1: header + both counts + `nLockTime`).
- Checked the Equihash solution length directly rather than through the helper — it's a byte length, not an element count, so the phrasing didn't fit.
- Documented that the constants deliberately count only one loop iteration and exclude per-element bytes held in later sections (a v5 spend carries another 192+64 in trailing vectors). The looseness is the point; tightening them risks rejecting valid input, and that's worth saying out loud so nobody "fixes" it later.

## Tests

Your cases all pin `count=1, remaining=0`, which never exercises the division, and nothing asserted the checks *accept* things. I added:

- a large-count case (`tx_out_count=1000` with 100 bytes left),
- `TestBoundsChecksAcceptMinimallySizedElements`, feeding each check exactly minimum-size elements — the regression these bounds could plausibly cause, and the one case that would catch a constant tightened by even a byte,
- both pool names for the action bundle.

`go test ./...` is green and `gofmt` is clean.

One note for the record, no change needed: the JoinSplit bound saves no memory — `joinSplit` is a zero-field struct, so `make([]joinSplit, N)` allocates nothing for any `N`. It only short-circuits the loop. Harmless, and it future-proofs the struct if those fields are ever uncommented.

LarryRuane (commented):
ACK 30a9f91ac40ff401c0bf8441e2cbc7484c130afa

LarryRuane (approved):
re-ACK 442c95ce2d4cf6d125633c46af9b3b93e3cfea81

PR #563, opened 2026-06-09
Detect reorgs that occur while the server is down

frankbraun:
## Summary

`NewBlockCache` rebuilds the block index from the on-disk cache but never initializes `latestHash` (the `setLatestHash` helper exists but is only called from `Reorg`). Since `HashMatch` vacuously accepts any block while `latestHash == hash32.Nil`, the first block ingested after **every** restart is appended without a prev-hash check against the cache tip.

Consequence: if the backing node reorgs across a lightwalletd restart, the new chain is silently appended on top of the orphaned blocks. The orphans remain in the append-only compact-block cache permanently, and the server hands out an internally inconsistent block stream (block N's `prevHash` doesn't match the block N−1 it actually serves).

## Real-world impact

This happened in production during the 2026-06-03 mainnet reorg: the zebrad instance backing zec.fyi followed a minority branch for heights 3364600–3364603 and was restarted while stuck on it; it came back on the canonical chain, and lightwalletd (restarted at the same time) appended canonical 3364604 directly on top of the orphaned 3364603 — no `REORG:` log line, no unwind. The server then served the four orphaned blocks, breaking prev-hash continuity at 3364604, until the cache was manually repaired a week later. Wallets that validate chain continuity stall on such a stream; wallets that don't can scan orphaned blocks.

## Fix

Call the existing `setLatestHash()` at the end of `NewBlockCache`. With `latestHash` restored from disk, the first ingest after a restart goes through the same `HashMatch` → `Reorg` walk-back as reorgs detected while running. A side benefit: when no reorg occurred, startup no longer drops and re-fetches the cache tip (previously, `getbestblockhash` could never match `Nil`, forcing an unnecessary `REORG: dropping block …` of the tip on every restart).

## Testing

- Added a regression test in the existing restart-simulation section of `TestCache`: after reopening the cache from disk, `latestHash` must equal the top block's hash and `HashMatch` must reject a non-connecting block. The test fails on master (`latestHash not initialized after restart`) and passes with this change.
- `go test ./...` passes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


LarryRuane:
ACK 8607db8adf797ef774ae3a3c04e8fd7d52a9c4f6
I verified the fix by stepping through the modified code with the debugger. This is a very safe fix. The only minor thing I noticed is that the comment above `setLatestHash()` says the caller should hold the mutex, but in this case, it doesn't.  But the lock isn't required, because the server is single-threaded during the startup phase. So the comment is wrong. I just pushed a second commit to fix this comment, and also added a note in `CHANGELOG.md`. I also verified that the new unit test code fails without the fix, as expected. Thanks for the fix!

Update: I was trying to merge this, but there was a conflict on CHANGELOG, which I resolved using the GitHub web interface (hadn't tried that before), which added another commit. (I don't think I would have preferred this, but it's okay I think, it works.)

PR #564, opened 2026-06-10
fix(common,frontend): make RawRequest context-aware for streaming RPC cancellation

ouicate:
## Summary

Addresses the residual goroutine/RPC-slot leak described in [GHSA-5h96-xw2v-jxgq](https://github.com/zcash/lightwalletd/security/advisories/GHSA-5h96-xw2v-jxgq) after PR #560.

PR #560 fixed the unbounded channel-send leak when a gRPC client disconnects mid-stream, but the producer could still remain blocked inside `getBlockFromRPC` → `RawRequest` for the duration of an in-flight zcashd JSON-RPC call. This change propagates `context.Context` through the full RPC path so cancelled streams abort in-flight HTTP requests promptly.

- Change `common.RawRequest` to accept `context.Context` as its first parameter.
- Thread `ctx` through `getBlockFromRPC`, `GetBlock`, and `GetBlockRange` (and gRPC handlers that call them).
- Add `frontend.NewContextRawRequest`, a context-aware HTTP POST JSON-RPC client using `http.NewRequestWithContext`. Production wiring replaces the previous `rpcClient.RawRequest` assignment. When btcd lands `RawRequestWithContext` ([btcsuite/btcd#2506](https://github.com/btcsuite/btcd/pull/2506)), this wrapper can delegate to it directly.
- Return immediately on `ctx.Err()` after `http.Client.Do` errors instead of retrying.

### DIFF
--- a/frontend/rawrequest.go
+++ b/frontend/rawrequest.go
@@ -1,0 +1,124 @@
+// Copyright (c) 2019-2020 The Zcash developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or https://www.opensource.org/licenses/mit-license.php .
+
+package frontend
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"net/http"
+	"time"
+
+	"github.com/btcsuite/btcd/btcjson"
+	"github.com/btcsuite/btcd/rpcclient"
+)
+
+const (
+	requestRetryInterval = 500 * time.Millisecond
+	defaultHTTPTimeout   = time.Minute
+)
+
+type jsonRPCResponse struct {
+	Result json.RawMessage   `json:"result"`
+	Error  *btcjson.RPCError `json:"error"`
+}
+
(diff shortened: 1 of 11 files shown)


ouicate:
Please take a look @nuttycom @LarryRuane @ebfull

frontend/rawrequest.go
```diff
@@ -0,0 +1,166 @@
+// Copyright (c) 2019-2020 The Zcash developers
```
nuttycom:
```suggestion
// Copyright (c) 2026 The Zcash developers
```

frontend/rawrequest_test.go
```diff
@@ -0,0 +1,114 @@
+// Copyright (c) 2019-2020 The Zcash developers
```
nuttycom:
```suggestion
// Copyright (c) 2026 The Zcash developers
```

nuttycom (approved):
utACK 87a0b37b5c046bdf2fbd21d4fbfbb4bd0310ef34

PR #573, opened 2026-07-27
Add support for the zakura backend

emersonian:
[Zakura](https://zakura.com) is a new Zcash node that lightwalletd is compatible with. Instead it currently refuses to start due to a version string check, which operators have no way of disabling.

This update recognizes the zakura node's subversion string alongside zcashd and zebrad, and adds a --no-backend-check option that skips backend detection entirely so lightwalletd can connect to any node that speaks the expected RPCs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,15 @@ and this library adheres to Rust's notion of
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 The most recent changes are listed first.
 
+## [Unreleased]
+
+### Added
+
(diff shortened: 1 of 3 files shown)


nuttycom:
@emersonian should we remove the support for zcashd?

LarryRuane (approved):
ACK looks good, reviewed, tested

emersonian:
I'm working on a separate PR to remove zcashd mentions, wanted to get this out as soon as possible with minimal review burden so I did not include it. @nuttycom 

Thanks for the quick merge.

Draft PR: #580

PR #574, opened 2026-07-27
Release v0.5.1: bound oversized client input (GHSA-q2c2, GHSA-6ppp, GHSA-4hp3)

LarryRuane:
Ports the v0.5.1 security fixes to master. Reviewed privately in `lightwalletd-security-fixes` (PRs #4, #5, and release PR #6); this is the public landing.

Fixes three advisories, all reported by **@ipwning**, all with the same shape — a client-supplied blob was expanded (hex-encoded, doubling it, then JSON-marshalled) and forwarded to the backend *before* any size check, so an unauthenticated client could force large allocations in lightwalletd and parsing work in the backend using input that could only ever be rejected:

| Advisory | Handler | Fix |
|---|---|---|
| **GHSA-q2c2-hpp9-58hm** | `GetTreeState` | reject a block hash that isn't 32 bytes |
| **GHSA-6ppp-r2gc-9q6v** | `SendTransaction` | reject a raw transaction over 2,000,000 bytes (the Zcash block size limit, above which a transaction is unminable by definition) |
| **GHSA-4hp3-3494-3f2m** | `GetMempoolTx` | cap the exclude list at 20,000 suffixes, and build it *before* taking the method mutex |

The reporter measured single 3.9 MB requests taking ~11.5 s instead of ~15 ms and adding ~30 MB of heap, with 12–16 concurrent requests OOM-killing a 256 MiB container; and a 120,000-suffix exclude list taking 1209 ms instead of 2 ms while returning zero transactions.

The `GetMempoolTx` change also moves the per-suffix work out of the critical section — it depends only on request data, so holding `s.mutex` across it let one caller's list stall every other `GetMempoolTx` caller.

Each guard rejects with `InvalidArgument`, matching the length check already present in `GetTransaction`.

## Tests

Every new test asserts both halves: that oversized input is rejected **and that zcashd is never contacted**, plus that the bound is not off by one —

- `TestGetTreeStateInvalidHashLength` — lengths 1/31/33/64/4 MiB rejected; a valid 32-byte hash still reaches the backend.
- `TestSendTransactionOversized` — `maxRawTxSize + 1` rejected; exactly `maxRawTxSize` still forwarded.
- `TestGetMempoolTxExcludeListCap` — over the cap rejected; exactly at the cap accepted.

Also extends the `resetGlobals` test helper to clear the mempool cache globals (`mempoolMap`, `mempoolList`, `lastMempool`), which otherwise leak between tests now that one refreshes the cache.

Full suite green under `-race -shuffle=on`; `gofmt` and `go vet` clean.

## CHANGELOG

Adds `## [0.5.1] - 2026-07-27`, covering these fixes plus the zakura backend support from #573. Everything from `[0.5.0]` down is unchanged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,17 @@ The most recent changes are listed first.
 
 ### Fixed
 
+- Validate the size of client-supplied blobs before expanding them.
+  `GetTreeState` now rejects a block hash that isn't 32 bytes
+  (GHSA-q2c2-hpp9-58hm), and `SendTransaction` rejects a raw transaction
+  larger than the 2,000,000-byte Zcash block size limit
+  (GHSA-6ppp-r2gc-9q6v). Both previously hex-encoded the bytes (doubling
+  them) and JSON-marshalled the result before forwarding to zcashd, so an
+  unauthenticated client could force large allocations in lightwalletd, and
+  parsing work in the backend, using input that could only ever be rejected —
+  enough in concurrent requests to drive the process into an out-of-memory
+  kill.
+
 - Bound the transparent-address gRPC methods against unbounded per-request
   work (GHSA-x4m7-3gpp-xc36). `GetTaddressBalanceStream` now caps the number of
   streamed addresses and validates each one as it arrives, instead of buffering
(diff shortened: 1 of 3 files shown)


LarryRuane (commented):
ACK

PR #583, opened 2026-07-30
Narrow the GetMempoolTx mutex to the cache it protects (GHSA-f9pw-q493-7qvh, GHSA-9p9r-mggr-8q9g)

LarryRuane:
Fixes **GHSA-f9pw-q493-7qvh** (reported by @dingledropper) and **GHSA-9p9r-mggr-8q9g** (reported by @Zk-nd3r), both Moderate. Developed and reviewed privately; this is the public landing.

## The problem

`GetMempoolTx` held `s.mutex` for its entire body, and two expensive things ran inside that critical section:

- **the mempool refresh** — one `getrawmempool` plus one `getrawtransaction` per newly-seen txid (GHSA-f9pw)
- **the response loop's `resp.Send` calls** — which block on the client's flow control (GHSA-9p9r)

So one caller's backend round-trips, or a single client that simply stopped reading its stream, stalled every other `GetMempoolTx` caller for as long as it took. The streaming case is unbounded, since the client decides when to read.

`GetMempoolTx` is currently the only method that takes `s.mutex`, so the stall was confined to its own concurrent callers rather than the whole gRPC service. That's a recent property rather than a design guarantee — `GetLatestBlock` held the same mutex until 5c25f20 removed it — and keeping the critical section down to the cache accesses is what keeps it true.

## The fix

The lock now covers only the shared cache. It's taken once to decide whether this call refreshes and to snapshot the current list and map, and again briefly to publish a new cache. The RPCs and the sends happen in between, unlocked.

Reading the snapshot without the lock is safe because a refresh never mutates the published map — it builds a new one and swaps the pointer — so a captured snapshot stays immutable for as long as it's used. Stamping `lastMempool` before releasing the mutex claims the refresh, so concurrent callers serve from the existing cache instead of all issuing the same RPCs.

**The refresh also now stops when the client that triggered it disconnects.** Since `RawRequest` observes the request context, without that check every remaining per-txid fetch fails once the client hangs up, each failure is swallowed by the existing "mempool transactions can disappear" path, and the handler publishes a cache missing nearly every transaction — degrading it for every other caller until the next refresh two seconds later.

Also fixes a smaller pre-existing wart: the refresh unmarshalled `getrawmempool` straight into the package-level txid list, so a refresh failing part-way left that list updated while the map still held the previous contents. Both are now built as locals and published together.

## Tests

Two, each verified to fail against the previous code rather than merely pass against the new one:

- `TestGetMempoolTxSlowClientDoesNotBlockOthers` parks one caller inside `Send` and asserts a second still completes. Against the old code it blocks for the full 10s timeout.
- `TestGetMempoolTxAbortsRefreshOnClientCancel` cancels mid-refresh and asserts the loop stops early without publishing a partial cache.

Full suite green under `-race -shuffle=on`; `gofmt` and `go vet` clean.

## Release

This PR also stages **v0.5.2**, which these fixes ship in — hence the `## [0.5.2] - 2026-07-30` CHANGELOG section rather than `[Unreleased]`.

The CHANGELOG is the only in-repo release artifact. There's no version constant to bump: the Makefile derives it at build time from ``git describe --tags`` (`-X .../common.Version=$(VERSION)`), so the tag *is* the version.

Remaining steps are all post-merge:

1. Tag `v0.5.2`
2. Create the GitHub release — note this is what ships the images: `docker_push` in `CI.yaml` runs `if: github.event_name == 'release'` and pushes `electriccoinco/lightwalletd:<tag>` and `:latest`, mirrored to `zodlinc`
3. Publish GHSA-f9pw-q493-7qvh and GHSA-9p9r-mggr-8q9g with `vulnerable <= 0.5.1` / `patched 0.5.2`

Naming the CHANGELOG section now rather than at tag time is deliberate: this repo has historically never renamed `[Unreleased]`, which is how it came to hold three releases' worth of entries by v0.5.0 (untangled there by splitting them across `[0.5.0]`, `[0.4.19]` and `[0.4.18]` based on which entries were present at each tag).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ and this library adheres to Rust's notion of
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 The most recent changes are listed first.
 
-## [Unreleased]
+## [0.5.2] - 2026-07-30
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,31 @@ and this library adheres to Rust's notion of
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 The most recent changes are listed first.
 
+## [Unreleased]
+
+### Fixed
+
+- Narrow the `GetMempoolTx` mutex to the cache snapshot it protects, instead
+  of holding it across the whole handler (GHSA-f9pw-q493-7qvh,
+  GHSA-9p9r-mggr-8q9g). The mempool refresh issues one `getrawmempool` plus a
+  `getrawtransaction` per new txid, and the response loop streams to the
+  client; both ran inside the critical section, so a single caller's backend
+  round-trips — or one client that simply stopped reading its stream — stalled
+  every other `GetMempoolTx` caller. The backend calls and the streaming sends
+  now run outside the lock, which is taken only to claim a refresh, snapshot
+  the cache, and publish the new one.
+
+- A `GetMempoolTx` refresh now stops when the client that triggered it
+  disconnects, instead of running to completion. Since the backend calls
+  observe the request context, every remaining fetch would otherwise fail and
+  be treated as a transaction that had left the mempool, publishing a cache
+  missing nearly all of its entries and degrading it for other callers until
+  the next refresh.
+
+- A `GetMempoolTx` refresh that fails part-way no longer leaves the cached
+  txid list updated while the transaction map still holds the previous
+  contents; the two are now published together.
+
 ## [0.5.1] - 2026-07-27
 
 ### Added
--- a/frontend/frontend_test.go
+++ b/frontend/frontend_test.go
@@ -1021,3 +1021,141 @@ func TestGetMempoolTxExcludeListCap(t *testing.T) {
 		t.Fatal("a request at the cap should have reached the backend")
 	}
 }
+
+// testmempoolslow is a GetMempoolTx stream whose Send blocks until released,
+// standing in for a client that has stopped reading its stream.
+type testmempoolslow struct {
+	walletrpc.CompactTxStreamer_GetMempoolTxServer
+	sending chan struct{}
+	release chan struct{}
+	once    bool
+}
+
+func (t *testmempoolslow) Context() context.Context { return context.Background() }
+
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,31 @@ and this library adheres to Rust's notion of
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 The most recent changes are listed first.
 
+## [0.5.2] - 2026-07-30
```
nullcopy, line 9: I'm not familiar with the release process on this project. Do we need to do anything else for a lightwalletd release other than set the version & release date in the CHANGELOG?

You might also mention this in the PR title/description

nullcopy (approved):
utACK e5233eda8b7121e2a6f91ec8ae1ce17f5d016858

CHANGELOG.md
LarryRuane: Good question — there's no documented release process for this repo (only an abandoned `docs/release-notes/release-notes-0.4.1.md`), so I dug into what it actually takes.

The CHANGELOG is the only in-repo artifact. There's no version constant to bump: the Makefile derives it at build time from `git describe --tags` (`-X .../common.Version=$(VERSION)`), so the tag *is* the version.

Everything else is post-merge:

1. Tag `v0.5.2`
2. Create the GitHub release — worth knowing this is what ships the images, not just bookkeeping: `docker_push` in `CI.yaml` runs `if: github.event_name == 'release'` and pushes `electriccoinco/lightwalletd:<tag>` and `:latest`, mirrored to `zodlinc`
3. Because these are security fixes, one step outside the repo: publish GHSA-f9pw-q493-7qvh and GHSA-9p9r-mggr-8q9g with `vulnerable <= 0.5.1` / `patched 0.5.2`

On naming the section `[0.5.2]` now rather than at tag time — that's deliberate. This repo has historically never renamed `[Unreleased]`, which is how it came to hold three releases' worth of entries by v0.5.0; that release untangled them by splitting across `[0.5.0]`, `[0.4.19]` and `[0.4.18]` according to which entries were present at each tag.

And agreed on the description — I've added a Release section covering the above. Probably worth writing this down in `docs/` at some point, since it's now been reconstructed from scratch twice.

PR #585, opened 2026-08-04
gRPC: guarantee block continuity in GetBlockRange

LarryRuane:
`GetBlockRange` resolves each height in the requested range independently, taking the block from the cache when it's present there and from the backend node otherwise. Those two sources agree except while the background ingestor is repairing a reorg: until it has rewound the cache, the cache holds blocks from the abandoned fork while the backend has already moved to the new one. A range spanning the cache tip could then return a block from the old fork followed by one from the new — a pair whose hashes don't link, describing a chain that never existed.

This tracks the hash that the next block has to present (going forward, the hash of the block just sent, which the next block must name as its parent; going backward, that block's parent, which the next block must be) and aborts the stream with `Aborted` rather than send a block that doesn't match.

`GetBlockRangeNullifiers` shares `common.GetBlockRange`, so it's covered too.

### Scope

This makes each *response* internally consistent. It does not prevent a range served entirely from the not-yet-repaired part of the cache from returning blocks from the abandoned fork — that resolves when the ingestor rewinds past them.

### Notes for review

- **Direction.** The reverse case (`Start` > `End`) is the easy one to get backwards. Both directions are covered by tests, and I checked by mutation that each fails if the comparison is written the wrong way round.
- **Darkside is unaffected.** `setPrevhash` (`common/darkside.go:254`) rewires prevhash across the whole active chain on every apply, so staged reorgs can't trip the check.
- **Retaining the hash across the loop is safe.** Only 32 bytes are kept, not the block; and `pruneCompactBlockToNullifiers` — the one consumer that mutates what it receives — touches only `Vtx` and `ChainMetadata`, never `Hash`/`PrevHash`.
- **New error path.** Clients will see a stream fail mid-range and must retry, which they already have to tolerate for network failures.

### Tests

Three tests in `common/common_test.go` after `TestGetBlockRange`, set up as the real thing rather than hand-fed to the checker: the cache holds a height-380640 block from an abandoned fork, the backend serves the genuine 380641 whose `PrevHash` names the *real* 380640.

- `TestGetBlockRangeDiscontinuity` — forward
- `TestGetBlockRangeDiscontinuityReverse` — same fixture, `Start` > `End`
- `TestGetBlockRangeContiguousReverse` — cached block really is the parent, so the range must stream normally; guards the happy path across the cache/backend boundary, which is every wallet syncing at the cache tip

Verified by mutation that the suite catches the check being disabled, either direction branch being dropped, and the direction flag being inverted.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,24 @@ and this library adheres to Rust's notion of
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 The most recent changes are listed first.
 
+## [0.5.3] - 2026-08-04
+
+### Fixed
+
+- `GetBlockRange` and `GetBlockRangeNullifiers` now check that each block they
+  return connects to the one before it, and fail the stream with `Aborted`
+  rather than send a block that doesn't (GHSA-m7j5-wvx3-qj6j). Every height in
+  the range is resolved independently, from the block cache when it's present
+  there and from the backend node otherwise, so while the background ingestor
+  was still repairing a reorg, a single request spanning the cache tip could
+  return a block from the abandoned fork followed by one from the current
+  fork. That pair violates `block[n+1].prevHash == block[n].hash` and
+  describes a chain that cannot exist, which a strict wallet will reject and a
+  lenient one may briefly index against. Note that this makes each response
+  internally consistent; a range served entirely from the not-yet-repaired
+  part of the cache still returns blocks from the abandoned fork, until the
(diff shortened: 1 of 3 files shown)


nullcopy (approved):
It took me a bit to wrap my head around all the direction reversals, but I got there 😅

utACK d0ca8f2db4e31c01443d8e32184918d8370072cb

PR #586, opened 2026-08-04
frontend: deduplicate client-supplied address lists

LarryRuane:
The transparent-address methods pass the client's address list to the backend unchanged, and the two backends treat repeats differently.

## `getaddressbalance` — a wrong number, not just a slow one

zcashd sums over the list entries, so a repeated address is counted repeatedly. Verified against zcashd master: the same address listed twice returned `100365254620` zats against `50182627310` listed once — exactly double.

zebrad parses the list into a `HashSet<Address>` before querying ([`methods.rs:3798`](https://github.com/ZcashFoundation/zebra/blob/main/zebra-rpc/src/methods.rs)), so duplicates collapse and it reports the balance of the distinct addresses.

So the same request returned different balances depending on which backend lightwalletd was configured with, and the zcashd answer was wrong. **This is a correctness fix, and it changes returned balances against a zcashd backend.**

It is not a security issue: a client can only mislead itself this way, not inflate anyone else's balance.

## `getaddressutxos` — a resource multiplier

Same split. zcashd looks up each entry independently and returned 250 UTXOs for an address listed twice that yields 125 listed once; zebrad returns the same set either way. Repeating an address costs the caller nothing — addresses are public, and neither a key nor any funds are needed to name one — so against a zcashd backend a single request could multiply the busiest address on the chain by `maxTaddrsPerRequest` (10,000).

zebrad was never exposed to this. The point of fixing it in lightwalletd is that lightwalletd shouldn't depend on backend behavior to bound work it can bound itself — the same layering argument as validating input locally.

## Details

- Both lists are deduplicated before the backend call, keeping first occurrence, so the order the backend sees is unchanged.
- For UTXOs the cap is applied to the raw list first, then dedupe, so the 10,000 limit remains a bound on input size and can't be slipped past with duplicates.
- The dup check runs before `checkTaddress`, so validation happens once per distinct address rather than once per entry. An invalid address is still rejected, since its first occurrence reaches the check.
- `GetTaddressTransactions` and `GetTaddressTxids` carry a single `address` field, so they can't receive duplicates and are unaffected.

## Scope

The UTXO change narrows the worst case; it does not close it. `StartHeight` and `MaxEntries` are still response filters applied after the entire backend result has been fetched and unmarshalled, so a single named address with a large UTXO set is bounded by nothing here — and that case is unaffected by deduplication. That residual is the unremediated part of GHSA-x4m7-3gpp-xc36 and needs range/limit arguments pushed down into the backend RPC to close.

## Testing

`TestGetTaddressBalanceDeduplicates` uses a stub that returns 1000 zats per list entry it receives — modelling zcashd — and asserts that four entries naming two distinct addresses yield 2000, not 4000.

`TestGetAddressUtxosDeduplicates` asserts the backend receives exactly the distinct addresses in first-occurrence order, and that an invalid address following duplicates still yields `InvalidArgument`.

Mutation-tested: bypassing either dedupe fails its test, and a variant that deduplicates without preserving order fails 8/8 runs (six addresses are used specifically so map iteration can't match by chance).

### DIFF
--- a/frontend/service.go
+++ b/frontend/service.go
@@ -586,14 +586,23 @@ func getTaddressBalanceZcashdRpc(ctx context.Context, addressList []string) (*wa
 		return nil, status.Errorf(codes.ResourceExhausted,
 			"getTaddressBalanceZcashdRpc: too many addresses (limit %d)", maxTaddrsPerRequest)
 	}
+	// Eliminate duplicate addresses; zcashd sums per list entry, so a repeated
+	// address would otherwise inflate the balance it reports.
+	addresses := make([]string, 0)
+	seen := make(map[string]struct{})
 	for _, addr := range addressList {
+		if _, dup := seen[addr]; dup {
+			continue
+		}
 		if err := checkTaddress(addr); err != nil {
 			return nil, err
 		}
+		seen[addr] = struct{}{}
+		addresses = append(addresses, addr)
 	}
 	params := make([]json.RawMessage, 1)
 	addrList := &common.ZcashdRpcRequestGetaddressbalance{
-		Addresses: addressList,
+		Addresses: addresses,
 	}
 	param, err := json.Marshal(addrList)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


frontend/service.go
```diff
@@ -886,22 +886,26 @@ func MempoolFilter(items, exclude []string) []string {
 }
 
 func getAddressUtxos(ctx context.Context, arg *walletrpc.GetAddressUtxosArg, f func(*walletrpc.GetAddressUtxosReply) error) error {
-	// Bound the address count before contacting zcashd: getaddressutxos cannot
-	// push down StartHeight/MaxEntries, so lightwalletd fetches and
-	// materializes the entire backend result before applying those limits.
-	// Capping the input keeps one request from forcing unbounded backend work
-	// and result materialization (GHSA-x4m7-3gpp-xc36).
+	// GHSA-x4m7-3gpp-xc36
```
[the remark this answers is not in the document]
LarryRuane: @CoPilot, I would prefer to keep comments to a minimum and not include concepts that are covered in the commit message. Comments only where something subtle or unexpected is being done.

LarryRuane:
@emersonian  - please review

LarryRuane:
Reviewed (though I am ostensibly the author, Claude wrote most of this); LGTM. I stepped through the new test, `TestGetTaddressBalanceDeduplicates`, in the debugger, and both the production and test code are working as expected.

frontend/service.go
```diff
@@ -886,22 +895,26 @@ func MempoolFilter(items, exclude []string) []string {
 }
 
 func getAddressUtxos(ctx context.Context, arg *walletrpc.GetAddressUtxosArg, f func(*walletrpc.GetAddressUtxosReply) error) error {
```
pacu, line 901: @LarryRuane is it possible to send a GetAddressUtxosArg with no addresses? I don't think the code is guarding from that directly, it does so indirectly when the dedup work is done. I think if that can happen we shall return an "invalid argument" reply of some sort

pacu (dismissed):
utACK modulo the comment resolution

frontend/service.go
LarryRuane: I don't think an empty address list should be considered an error, it shouldn't be a special case, and I verified that we don't return an error (either before or after de-dup). Since you asked about it, I (Claude) added tests (force-pushed) to make sure this case doesn't return an error. Zebrad also doesn't mind an empty list (I tried it), it didn't fail. I think this is now ready to merge.

pacu (approved):

PR #589, opened 2026-08-10
common: report lightwalletProtocolVersion in GetLightdInfo

emersonian:
LightdInfo.lightwalletProtocolVersion is not populated. It is declared in the generated walletrpc bindings and read by clients, but not served by lightwalletd.

The field is specified in lightwallet-protocol at lightwallet-protocol/walletrpc/service.proto:118:

    string lightwalletProtocolVersion = 18; // version of
    // https://github.com/zcash/lightwallet-protocol served by this server

and the same specification makes checking it a client requirement, in the documentation of BlockRange at service.proto:35-38:

    Clients MUST verify that the version of the server they are
    connected to are capable of returning pruned and/or transparent
    data before setting `poolTypes` to a non-empty value.

Both were introduced together in lightwallet-protocol v0.4.0, which is what ties them: the version string is the means by which a client discharges that MUST. 

**The consequence of leaving the field empty is that it is indistinguishable from a pre-v0.4.0 server.**

A client that honors the MUST has no way to tell a server that serves transparent and Ironwood data inside compact blocks from one that does not, so it must fall back to shielded-only scanning, against servers that have in fact served the data since 0.5.0.

Confirmed against a public 0.5.2 deployment: GetBlockRange with poolTypes set returns vin, vout and ironwoodActions exactly as specified, while GetLightdInfo on the same server reports no protocol version at all.

The practical result is that transparent light-wallet support looks unavailable everywhere, and clients that want it have to be told out of band, by the operator, to disregard the probe.

### Implementation Note

This reports the version as a constant rather than one of the build-stamped variables above it: it describes the protocol served, not the build, and letting 'make build' overwrite it would defeat the point. It must be updated together with the lightwallet-protocol subtree, and only once the server actually serves what the named version specifies. The subtree is currently at v0.5.0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,22 @@ and this library adheres to Rust's notion of
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 The most recent changes are listed first.
 
+## [Unreleased]
+
+### Fixed
+
+- Report `lightwalletProtocolVersion` in `GetLightdInfo`. The field was added
+  to `LightdInfo` in lightwallet-protocol v0.4.0, in the same release as
+  `BlockRange.poolTypes`, and is the signal clients are required to check
+  before requesting non-default pool types. lightwalletd never populated it,
+  so it was always the empty string, and a client following that rule could
+  not tell a server that serves transparent and Ironwood data inside compact
+  blocks from one that does not. Correctly-behaving clients therefore had to
+  fall back to shielded-only scanning, or rely on an out-of-band assertion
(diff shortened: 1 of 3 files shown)


emersonian:
Worth noting that Zaino populates this field. Currently zecd users cannot connect to lightwalletd if they want T-address support, Zec.rocks is running a fork that has this patch.

https://github.com/zingolabs/zaino/blob/f8b99861ce48ce51823adab8db7605994356aef0/packages/zaino-state/src/indexer/node_backed_indexer.rs#L2062

pacu (approved):
utACK

Thank you @emersonian

PR #590, opened 2026-08-14
frontend: apply the address-count cap to GetTaddressBalance

LarryRuane:
`maxTaddrsPerRequest` (10,000) bounds `GetTaddressBalanceStream`, `GetAddressUtxos` and `GetAddressUtxosStream`, but not the unary `GetTaddressBalance`.

The comment on the constant claimed that method was "already implicitly bounded by gRPC's `MaxRecvMsgSize`", giving "an equivalent bound". It is bounded — but not equivalently. The unary request carries its whole list in a single protobuf message, so grpc-go's 4MB default applies to the message as a whole:

```
4,194,304 bytes / 37 bytes per address ≈ 113,000 addresses
```

(37 = 1 tag byte + 1 length byte + 35 address characters.) That is eleven times the cap every sibling method enforces.

The streaming method genuinely does need its own explicit cap, because each `Address` arrives as its own small message and `MaxRecvMsgSize` never applies to the total — that part of the original reasoning was right. It just doesn't follow that the unary bound is equivalent.

## Impact

One method forwards an order of magnitude more addresses to the backend than any other allows, and returns a single `int64`. Small reply, large backend work — up to ~113,000 address-index lookups per request.

The 4MB request is real cost to the attacker, so this is weaker than an unbounded vector, and it is not the worst amplification path in this area. But there is no reason for one of four methods taking the same input to have a bound an order of magnitude looser than the others, discoverable only by dividing 4MB by 37.

The deduplication in #586 does not help here: distinct addresses cost the attacker nothing to generate, and they need not be funded (they do need valid base58 checksums, since `checkTaddress` only checks the character set — but those are trivially generated offline).

## Change

The cap goes in `getTaddressBalanceZcashdRpc`, the choke point both the unary and streaming paths call, so it covers both. The stream keeps its own check, which rejects addresses as they arrive rather than after buffering them.

The stale claim is dropped from the constant's comment.

## Testing

`TestGetTaddressBalanceTooManyAddresses` asserts that `maxTaddrsPerRequest + 1` addresses yield `ResourceExhausted` and that the backend is never contacted. Mutation-tested: removing the cap fails the test.

Note this PR touches `getTaddressBalanceZcashdRpc`, as does #586. Whichever merges second may need a trivial conflict resolution — the two changes are independent (one bounds the list, the other deduplicates it).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,15 @@ The most recent changes are listed first.
 
 ### Fixed
 
+- `GetTaddressBalance` now rejects an address list longer than the same 10,000
+  limit that `GetTaddressBalanceStream`, `GetAddressUtxos` and
+  `GetAddressUtxosStream` already enforce. The unary method takes its whole
+  list in one protobuf message, so its only bound was gRPC's 4MB
+  `MaxRecvMsgSize` — roughly 113,000 addresses at 37 wire bytes each, an order
+  of magnitude above the intended cap. A single request could therefore ask
+  the backend for eleven times more address-index lookups than any other
+  method allows, and return only an 8-byte balance (GHSA-x4m7-3gpp-xc36).
+
 - Report `lightwalletProtocolVersion` in `GetLightdInfo`. The field was added
   to `LightdInfo` in lightwallet-protocol v0.4.0, in the same release as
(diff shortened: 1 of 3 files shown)


pacu (approved):

pacu:
utACK 

Nice catch! thank you @LarryRuane !

PR #595, opened 2026-08-25
frontend: reject an unknown shielded protocol with InvalidArgument

pucedoteth:
## The bug

`GetSubtreeRoots` validates `arg.ShieldedProtocol` and then returns a bare Go error:

```go
default:
    return errors.New("unrecognized shielded protocol")
```

gRPC has no code to read off a bare error, so it reaches the client as `codes.Unknown` — *"the server had a problem"*. That is a code clients back off and retry on, and this request cannot succeed until the caller changes it. The right answer is `InvalidArgument`: the caller's to fix, don't retry.

Every other rejection of client input in this file returns a status error — including the `json.Marshal` check four lines further down in the same function. These were the only two bare ones, out of 81 error returns in `frontend/service.go`.

Measured against the handler:

```
status.Code(err) == codes.Unknown          // before
status.Code(err) == codes.InvalidArgument  // after
```

The darkside `Reset` handler's `ChainName` check had the same shape, four lines below its `BranchID` check, which already used `InvalidArgument`. Fixed alongside, since it is the same one-line pattern in the same file.

## Tests

`GetSubtreeRoots` had no test at all. The new one pins the code for an unrecognized protocol and asserts the backend is never contacted, then checks that all three recognized protocols still reach `z_getsubtreesbyindex` under the names it expects — so the switch cannot be tightened into rejecting a valid protocol without the test noticing.

Reverting only `frontend/service.go` and keeping the test fails as:

```
frontend_test.go:1473: expected InvalidArgument for an unrecognized shielded protocol,
    got: Unknown unrecognized shielded protocol
```

`go test ./...` is green across all seven packages, `go vet ./frontend/` is clean, and `gofmt -l frontend/` is empty. (`go vet ./testclient/` reports three unexported-json-tag warnings on master already; untouched here.)

Changelog entry added under Unreleased → Fixed.

## Scope note

Removing the last two uses of `errors.New` in this file left the `errors` import unused, so it goes too — that is the only reason the import block is in the diff.

---
🤖 Written with [Claude Code](https://claude.com/claude-code). Every result above is from a local run.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,14 @@ The most recent changes are listed first.
 
 ### Fixed
 
+- `GetSubtreeRoots` now rejects an unrecognized `ShieldedProtocol` with
+  `InvalidArgument` instead of `Unknown`. It was the one production handler
+  returning a bare Go error, which gRPC surfaces as `codes.Unknown` -- the code
+  a client reads as "the server had a problem" and backs off and retries on,
+  for a request that cannot succeed until the caller changes it. The darkside
+  `Reset` handler's `ChainName` check returned a bare error the same way, four
+  lines below its `BranchID` check, which already used `InvalidArgument`.
+
 - `GetTaddressBalance` now rejects an address list longer than the same 10,000
   limit that `GetTaddressBalanceStream`, `GetAddressUtxos` and
   `GetAddressUtxosStream` already enforce. The unary method takes its whole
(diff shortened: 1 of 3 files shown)


LarryRuane (approved):
ACK 7170df2499849ec04da30c77a6e8bd6b5ddf3329
Thanks, good fix!

PR #596, opened 2026-08-27
docs: add public Zebra endpoint benchmark link in Zebrad section

Flotapponnier:
When setting up lightwalletd, developers often need a quick public Zebra endpoint to test against before their own node finishes syncing. The Zebrad section already explains the RPC interface and which methods are used — adding a pointer to a live benchmark of available keyless endpoints is a natural next step.

OpenChainBench probes Tatum Zebra, Tatum zcashd, and Blockchair every 60 seconds from 3 regions and surfaces p50/p90/p99 latency: https://openchainbench.com/benchmarks/zcash-rpc

Placement: right after the list of used RPCs, before the Lightwalletd build section.


LarryRuane:
Thanks for the PR. We're going to pass on this one.

lightwalletd issues address-level RPCs (`getaddressbalance`, `getaddressutxos`) and broadcasts transactions through its backend node. Pointing it at a third-party public endpoint hands that provider every address its users look up and every transaction they send. We don't want the README recommending that configuration, even as a temporary measure — lightwalletd's whole purpose is to keep that information from leaking, and "temporary" test setups have a way of becoming permanent.

We also don't link to third-party commercial services from the README generally.

PR #599, opened 2026-08-27
CI: fix the zodlinc mirror step, which has never succeeded

LarryRuane:
The `Push to zodlinc` step in `docker_push` has never succeeded. It failed at v0.5.4 with:

```
Error: unknown flag: --src-creds
```

`crane copy` has no `--src-creds` or `--dst-creds`. Those are skopeo's flags, and skopeo spells the second one `--dest-creds`, so the step was written against a different tool's interface.

It also failed at v0.5.2 and v0.5.3, there with `crane: command not found` — `go install` writes to `GOPATH/bin`, which is not on `PATH` in this job. That was fixed after v0.5.3, and fixing it is what let the run get far enough to hit the flags. The net effect is that the mirror has been updated by hand after every release, v0.5.4 included: `zodlinc/lightwalletd:v0.5.4` and `:latest` were copied across by hand on 2026-08-28, the day after the release.

The credentials the step was trying to pass turn out not to be needed. The only image it reads is `electriccoinco/lightwalletd:<tag>`, and that repository is public, so the docker login session conflict described in the old comment does not arise: authenticate as the destination org, then copy. The two source secrets leave the step's environment along with the flags.

**This cannot be verified before merge.** `docker_push` runs only on `if: github.event_name == 'release'`, so the first real exercise is the next release publish. Worth watching that run rather than assuming it.


LarryRuane:
@y4ssi friendly ping on this one — it's CI-only, green, and has no conflicts. The mirror step still has never succeeded, so v0.5.4 had to be copied across by hand again on 08-28; this is the change that stops that happening every release. Happy to hand it to someone else if you're stretched.

LarryRuane:
Since this can't be exercised before merge, here are its three claims checked directly against `crane` v0.21.8 — the same version the step installs. None of this needs any Actions knowledge; the commands can be pasted as-is.

**1. `crane copy` has no `--src-creds` or `--dst-creds`.**

```
$ go run github.com/google/go-containerregistry/cmd/crane@v0.21.8 copy --help
Flags:
  -a, --all-tags     (Optional) if true, copy all tags from SRC to DST
  -h, --help         help for copy
  -j, --jobs int     (Optional) The maximum number of concurrent copies, defaults to GOMAXPROCS
  -n, --no-clobber   (Optional) if true, avoid overwriting existing tags in DST
```

That is the complete flag set, which is where `Error: unknown flag: --src-creds` came from.

**2. `crane auth login` exists and takes `-u` / `-p`, as this PR uses it.**

```
$ go run github.com/google/go-containerregistry/cmd/crane@v0.21.8 auth login --help
Flags:
  -h, --help              help for login
  -p, --password string   Password
      --password-stdin    Take the password from stdin
  -u, --username string   Username
```

**3. The source repository is public, so it needs no credentials.**

```
$ go run github.com/google/go-containerregistry/cmd/crane@v0.21.8 digest electriccoinco/lightwalletd:v0.5.4
sha256:2ae3a551e1117eea17e9742b1360fc36db0e6d0613ddcbc7f5eed68a9b34e3db
```

That ran with no Docker Hub credentials at all, and the digest matches what `zodlinc/lightwalletd:v0.5.4` and `:latest` currently point at — so dropping `SRC_USER`/`SRC_PASS` costs nothing.

What this does not cover: the `ZODLINC_DOCKERHUB_*` secrets and the destination push permission, neither of which can be exercised outside the workflow. `docker_push` is gated on `if: github.event_name == 'release'`, so the first real run is still the next release publish.

pacu (approved):

y4ssi (approved):
utACK

